PR #466, opened 2023-09-05
[Certora] Via IR

QGarchery:


QGarchery:
Closing as not planned because it requires to munge the safe transfer lib. This would add the risk to do a mistake in the process, and hide some bugs because of it. Trusting the compiler seems like a safer option

PR #522, opened 2023-10-08
Healthiness preservation

QGarchery:
HEAVILY inspired from [this gist](https://gist.github.com/daejunpark/6f139d12d025aa80732846945e0ee900), thanks @daejunpark !


QGarchery:
Replaced by #598 because it's simpler and it gives an computation to ensure that your loan is safe for a period `elapsed`

PR #530, opened 2023-10-10
Fix gas cost study for liquidation?

MerlinEgalite:
Adding `randomForwardTimestamp` to have cold slots before the liquidation.

## `closePosition = false`:

Screenshot 2023-10-10 at 17 47 13

## `closePosition = true`:

Screenshot 2023-10-10 at 17 48 56

## Question

Should I change the comment btw? "// Realize the bad debt if needed. Note that it saves ~3k gas to do it."

### DIFF
--- a/test/forge/integration/LiquidateGasCostIntegrationTest.sol
+++ b/test/forge/integration/LiquidateGasCostIntegrationTest.sol
@@ -1,0 +1,59 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+import "../BaseTest.sol";
+
+contract LiquidateGasCostIntegrationTest is BaseTest {
+    using MathLib for uint256;
+    using MorphoLib for IMorpho;
+    using SharesMathLib for uint256;
+
+    uint256 amountSupplied;
+    uint256 amountCollateral;
+    uint256 amountBorrowed;
+    uint256 priceCollateral;
+    uint256 sharesRepaid;
+
+    function setUp() public override {
+        super.setUp();
+
+        (amountCollateral, amountBorrowed, priceCollateral) =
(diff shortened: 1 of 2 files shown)


QGarchery (approved):
Yes, we should change the comment, either:
- ~3k -> ~2k
- or something more vague, because this is a rough estimate. Maybe just "saves gas overall".

MerlinEgalite:
With simple foundry tests, here is what I get:

## Don't realize bad debt
Screenshot 2023-10-10 at 19 11 18

## Realize bad debt
Screenshot 2023-10-10 at 19 11 32

QGarchery:
I looked at the wrong column in your hardhat tests, I realize that it no longer saves a few thousand gas anymore

MerlinEgalite:
> I looked at the wrong column in your hardhat tests, I realize that it no longer saves a few thousand gas anymore

It depends if you check hardhat or foundry...

MathisGD:
> It depends if you check hardhat or foundry...

I think that both foundry and hardhat show that the improvement is no longer there

MerlinEgalite:
> I think that both foundry and hardhat show that the improvement is no longer there

Not foundry if I'm reading well

MathisGD:
don't realize image

realize image

MerlinEgalite:
> don't realize image
> 
> realize image

yes for hardhat but I'm talking about foundry

MathisGD:
oops indeed

MathisGD:
but the difference in the foundry test seems too big, it seems wrong to me

MerlinEgalite:
> but the difference in the foundry test seems too big, it seems wrong to me

Which of the test seems wrong to you hardhat or foundry?

MathisGD:
foundry seems wrong

Issue #535, opened 2023-10-11
Oracle integration is confusing and forces unnecessary deployments


alcueca:
This issue has two parts, but is submitted as one as they are both closely related.

Prices from the oracle are retrieved as `IOracle(marketParams.oracle).price()`. The lack of parameters in `price` means that there must be a separate contract for each possible loan/collateral pair. Furthermore, it is not clear from the code if the price is loan/collateral or collateral/loan.

I would suggest that at least you use `IOracle(marketParams.oracle).price(baseToken, quoteToken)` so that oracle aggregators can be used, and so that it is clearer that you are getting the price of one unit of `baseToken` in `quoteToken` terms.

In the code this would be `IOracle(marketParams.oracle).price(marketParams.collateralToken, marketParams.loanToken)` when you want the price of one collateral in loan terms, and `IOracle(marketParams.oracle).price(marketParams.loanToken, marketParams.collateralToken)` when you want the price of one loan token in collateral terms.

---

Going further, I would suggest that you think in terms of value, instead of price, and use a function signature such as `function value(address base, address quote, uint256 baseAmount) external view returns (uint256 quoteAmount);` so that the code is clearer. For example, this is how you would calculate the value of the collateral in loan terms for `_isHealthy`:

```
        uint256 collateralValue = IOracle(marketParams.oracle).value(
            marketParams.collateralToken,
            marketParams.loanToken,
            uint256(position[id][borrower].collateral
        ));
        uint256 maxBorrow = collateralValue.wMulDown(marketParams.lltv);
```

If adopting the last recommendation, `liquidate` would have to be refactored to calculate `seizedAssets` or `repaidShares` as position proportions, to avoid a second call to the oracle.

MerlinEgalite:
## Part 1

There this a comment specifying what `price()` should return in the `IOracle` file:
https://github.com/morpho-org/morpho-blue/blob/6be8b4c978249e4896aeb271128340b6d8bc7f6f/src/interfaces/IOracle.sol#L9-L12

Perhaps we should make it clearer?

## Part 2

That would mean passing another argument to the oracle while it's not necessary currently. 

I need to check the extent of the refactor.

alcueca:
On part 1, the comments are clear, but the code is not, and code and comments are in different files. Besides, other people less familiar than you will build oracle adaptors, and their failures might reflect bad on you. I would just take the extra step of the two asset parameters in `price` as a way of solving one issue, and preventing another.

[...]

PR #540, opened 2023-10-13
Use asset values instead of asset prices in oracle calls

alcueca:
Refactor to integrate with oracles using asset value terms instead of asset price terms.

Something is off in the liquidations, needs some debugging.

Also, the liquidations now do two oracle calls (one for `_isHealthy` and another to determine the liquidation), that is probably a non-starter due to gas reasons. If the liquidations can be refactored to work on position proportions instead of price then we wouldn't need the second oracle call.


MathisGD:
Thank you for your POC. I'm not in favor of continuing its development, as I don't find that it makes the code so much clearer. Also, as you mentioned in #535, we loose an optimization avoiding to call the oracle twice.

One advantage that it has, that I have not see mentioned, is that it enables complex oracles returning a different price depending on the amount (to reflect the market). But it's not even sure that such oracle will ever exist.

PR #541, opened 2023-10-13
feat(audit): fixing typo in security review report

tomrpl:
Updating the security review report


MathisGD:
isn't it better to keep the same date?

MerlinEgalite (changes requested):
Same question, why changing the date?

Rubilmax (approved):

Rubilmax:
> @tomrpl yes original date (we don't really care about the date of the edit)

the date inside the doc has been updated

MerlinEgalite (approved):

Rubilmax:
> ok but if we find an other typo in 6 month we don't want to update the date right. What we care about is the date of the audit

I agree, but then we need to warn OZ not to update the date of the report?

tomrpl:
> then we need to warn OZ not to update the date of the report?
In my opinion this is the best way to proceed!

MathisGD (approved):

PR #542, opened 2023-10-13
Remove bad debt realization mechanism

MathisGD:
From the whitepaper:

> Morpho Blue takes a different approach by accounting for the bad debt. When a liquidation occurs on Morpho Blue, if the borrower has outstanding debt but no collateral, the losses are socialized proportionately among lenders, resulting in an instant loss for lenders.
> 

The goals of this feature are:

1. avoid bank run where people either loose nothing or loose everything (the one that didn’t ran away).
2. remove the idea that a market can be fundamentally riskier and riskier become some bad debt is accumulating.

### Lenders can avoid haircut

Nothing ensures that liquidators realize the bad debts, even if incentives tend to make them do it. It can happen that they don’t liquidate positions to the nearest wei, on purpose or not. In this situation, lenders who notice the situation can withdraw their funds, realize the bad debt and resupply their funds. This way, they don’t occur any loss, but other lenders occur more losses.

Lenders could also well become liquidators themselves to do this “manipulation” before liquidators who might do it. Even if it is a special role requiring particular skills and infrastructure, they technically have more incentives to do it than liquidators and this has to be taken into account. 

These issues were known and acknowledged (but still reported by OZ in [[their audit](https://github.com/morpho-org/morpho-blue/blob/main/audits/2023-09-27-morpho-blue-and-speed-jump-irm-open-zeppelin.pdf)](https://github.com/morpho-org/morpho-blue/blob/main/audits/2023-09-27-morpho-blue-and-speed-jump-irm-open-zeppelin.pdf) of Morpho Blue). As a result, bad debt realization is not completely fair (some people can loose more than other).

### Lenders can pay liquidators to avoid haircut

There is actually a deeper issue: lenders could well give authorization to manage their position to a special contract that allows liquidators to do this manipulations on behalf of one of multiple lenders. The lenders incentivize the liquidator to do so by sharing with them the revenue (or the non loss) of the operation.

With this mechanism, multiple lenders can now profit from the “manipulation”, and they all have incentives to try using it. The equilibrium of this “game” is that the lenders who are not subscribing to such a contract, or not incentivizing enough the liquidator take all the bad debt for themselves (potentially occurring big relative loss, up to 100% of their funds). Even if the idea is very theoretical, the incentives pull the system to something that can even be seen as worse than before (1. happens for each new bad debt, instead of only when the bad debt becomes too big, but on smaller amounts).

Big thanks to @Picodes from Angle Labs who highlighted this second issue.

### The solution

The solution is to remove the bad debt realization mechanism. It seems that there is no good way to solve 1. in the context of a pool like Morpho Blue markets. And 2. is mitigated by the idea that you can recreate the market when you are no longer confortable with the bad debt that it contains.


src/Morpho.sol
```diff
@@ -373,21 +373,10 @@ contract Morpho is IMorpho {
 
         position[id][borrower].collateral -= seizedAssets.toUint128();
 
-        // Realize the bad debt if needed. Note that it saves ~3k gas to do it.
```
MerlinEgalite, line 376: 😢

MerlinEgalite (commented):
Maybe we should update the WPP at the same time?

MathisGD:
>Maybe we should update the WPP at the same time?

yes, I was waiting for the reviews of the protocol team on overleaf before pushing it here

MathisGD:
Finally, we decided to not remove the bad debt realization mechanism. Here are the reasons:
- first it's important to realize that this decision isn't primordial, in the sense that bad debt arises when lenders "lost" and it's not part of the normal operation of a market (at least not in considerable amounts). So realizing or not the bad debt doesn't change Morpho Blue's value.
- both the classic bank run scenario and the bad debt realization "problem" described above show unfair outcomes that seem non solvable. It's unclear what's best for lenders and under which conditions (studied deeper, we can share more insight if someone is interested).
- the bad debt realization offers the great property of cleaning the little bad debts coming to markets (of small accounts notably), and in this sense "markets run forever", which is a good property in lender's mind.

PR #547, opened 2023-10-17
build(npm): use bun

Rubilmax:


julien-devatom (commented):
I've not tested enough bun to have a concrete opinion on it.

test/hardhat/Morpho.spec.ts
```diff
@@ -22,7 +22,7 @@ const random = () => {
 const identifier = (marketParams: MarketParamsStruct) => {
   const encodedMarket = AbiCoder.defaultAbiCoder().encode(
     ["address", "address", "address", "address", "uint256"],
-    Object.values(marketParams),
+    Object.values(marketParams)
```
Rubilmax, line 25: I had to downgrade prettier I don't know yet why

MathisGD:
I'm not sure that we should switch to bun, especially in this repo. The performance improvement is cool but we're not sure about the stability and Morpho Labs' devs are much more used to yarn

Rubilmax:
> Morpho Labs' devs are much more used to yarn

If we don't switch to `bun` Morpho Labs devs will never be used to it :D

MathisGD:
Yes good remark, but we could do it elsewhere ^^

PR #548, opened 2023-10-17
Rename `IMorphoMarketStruct` to `IMorphoMarketGetter`

MerlinEgalite:
Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/51

### DIFF
--- a/src/libraries/periphery/MorphoBalancesLib.sol
+++ b/src/libraries/periphery/MorphoBalancesLib.sol
@@ -10,7 +10,7 @@ import {MorphoLib} from "./MorphoLib.sol";
 import {SharesMathLib} from "../SharesMathLib.sol";
 import {MarketParamsLib} from "../MarketParamsLib.sol";
 
-interface IMorphoMarketStruct {
+interface IMorphoMarketGetter {
     function market(Id id) external view returns (Market memory);
 }
 
@@ -41,7 +41,7 @@ library MorphoBalancesLib {
     {
         Id id = marketParams.id();
 
-        Market memory market = IMorphoMarketStruct(address(morpho)).market(id);
+        Market memory market = IMorphoMarketGetter(address(morpho)).market(id);
 
         uint256 elapsed = block.timestamp - market.lastUpdate;
 


Rubilmax (approved):

Jean-Grimal (approved):

src/libraries/periphery/MorphoBalancesLib.sol
```diff
@@ -10,7 +10,7 @@ import {MorphoLib} from "./MorphoLib.sol";
 import {SharesMathLib} from "../SharesMathLib.sol";
 import {MarketParamsLib} from "../MarketParamsLib.sol";
 
-interface IMorphoMarketStruct {
+interface IMorphoMarketGetter {
```
MathisGD, line 13:
```suggestion
interface IMorpho {
```

MerlinEgalite: I don't think it's good for integrators as it would create a clash if they to use both interface...

MathisGD: they could rename it, but as you want

Rubilmax: > note that this
> 
> ```solidity
> Market memory market = IMorphoMarketGetter(address(morpho)).market(id);
> ```
> 
> becomes
> 
> ```solidity
> Market memory market = morpho.market(id);
> ```

What you suggest forces the integrator to use the `IMorpho` defined locally instead of the `IMorpho` from `IMorpho.sol`. I am strongly against: what's useful with these libraries is that they are plug-and-play with the standard `IMorpho` interface

MathisGD: Nice catch

QGarchery (changes requested):
I really think we should do something similar to https://github.com/morpho-org/metamorpho/pull/240, which means that it would replace this PR.

I've already had to do some tricks manually multiple times and it makes the DX worse for no benefit IMO

MerlinEgalite:
> > I really think we should do something similar to [morpho-org/metamorpho#240](https://github.com/morpho-org/metamorpho/pull/240), which means that it would replace this PR.
> > I've already had to do some tricks manually multiple times and it makes the DX worse for no benefit IMO
> 
> Should I do a PR on top?

Done in #567

MathisGD:
To close right ? (replaced by #567)

MerlinEgalite:
#567 was supposed to be merged in this PR so I don't get why we're closing this one

Rubilmax:
> #567 was supposed to be merged in this PR so I don't get why we're closing this one

We could shortcut to `cantina-review` instead (or merge it now)

MerlinEgalite:
Indeed, we can do that. I agree to close this one then.

PR #549, opened 2023-10-17
Emit owner at deployment

MerlinEgalite:
Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/34

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -67,8 +67,10 @@ contract Morpho is IMorpho {
     constructor(address newOwner) {
         require(newOwner != address(0), ErrorsLib.ZERO_ADDRESS);
 
-        owner = newOwner;
         DOMAIN_SEPARATOR = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
+        owner = newOwner;
+
+        emit EventsLib.SetOwner(newOwner);
     }
 
     /* MODIFIERS */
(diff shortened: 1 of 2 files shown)


src/Morpho.sol
```diff
@@ -67,8 +67,10 @@ contract Morpho is IMorpho {
     constructor(address newOwner) {
         require(newOwner != address(0), ErrorsLib.ZERO_ADDRESS);
 
-        owner = newOwner;
         DOMAIN_SEPARATOR = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
+        owner = newOwner;
+
+        emit EventsLib.SetOwner(newOwner);
```
Rubilmax, line 73: I would internalize it in a function... But @MathisGD @QGarchery will be against?

Rubilmax (approved):
Internalizing the storage update and the event emitted would have prevented this, the same way using a third-party `Ownable` contract would have... I don't get why we're this stubborn

Jean-Grimal (approved):

MerlinEgalite:
@QGarchery can you have a look please?

QGarchery:
> I don't get why we're this stubborn

It could be said that suggesting the same thing multiple times is the subborn behavior :wink: 

The refactor you are suggesting can lead to unexpected side-effects too, like in [this issue](https://github.com/morpho-org/metamorpho/issues/221). There are trade-offs to different implementations, which fits different needs, but let's go into the details again

PR #550, opened 2023-10-17
Rename expected balance for assets

MerlinEgalite:
- Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/52

### DIFF
--- a/src/libraries/periphery/MorphoBalancesLib.sol
+++ b/src/libraries/periphery/MorphoBalancesLib.sol
@@ -91,7 +91,7 @@ library MorphoBalancesLib {
         (, totalSupplyShares,,) = expectedMarketBalances(morpho, marketParams);
     }
 
-    /// @notice Returns the expected supply assets balance of a `user` on a market after having accrued interest.
+    /// @notice Returns the expected supply assets balance of `user` on a market after having accrued interest.
     /// @dev Warning: Wrong for `feeRecipient` because their supply shares increase is not taken into account.
     function expectedSupplyAssets(IMorpho morpho, MarketParams memory marketParams, address user)
         internal
@@ -105,7 +105,7 @@ library MorphoBalancesLib {
         return supplyShares.toAssetsDown(totalSupplyAssets, totalSupplyShares);
     }
 
-    /// @notice Returns the expected borrow assets balance of a `user` on a market after having accrued interest.
+    /// @notice Returns the expected borrow assets balance of `user` on a market after having accrued interest.
     function expectedBorrowAssets(IMorpho morpho, MarketParams memory marketParams, address user)
         internal
         view
--- a/src/libraries/periphery/MorphoBalancesLib.sol
+++ b/src/libraries/periphery/MorphoBalancesLib.sol
@@ -91,7 +91,7 @@ library MorphoBalancesLib {
         (, totalSupplyShares,,) = expectedMarketBalances(morpho, marketParams);
     }
 
-    /// @notice Returns the expected supply balance of a user on a market after having accrued interest.
+    /// @notice Returns the expected supply assets balance of a `user` on a market after having accrued interest.
     /// @dev Warning: Wrong for `feeRecipient` because their supply shares increase is not taken into account.
     function expectedSupplyAssets(IMorpho morpho, MarketParams memory marketParams, address user)
         internal
@@ -105,7 +105,7 @@ library MorphoBalancesLib {
         return supplyShares.toAssetsDown(totalSupplyAssets, totalSupplyShares);
     }
 
-    /// @notice Returns the expected borrow balance of a user on a market after having accrued interest.
+    /// @notice Returns the expected borrow assets balance of a `user` on a market after having accrued interest.
     function expectedBorrowAssets(IMorpho morpho, MarketParams memory marketParams, address user)
         internal
         view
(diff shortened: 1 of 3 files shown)


Rubilmax (approved):

Jean-Grimal (dismissed):

MathisGD (dismissed):

julien-devatom (commented):
Why `expected` and not `current`? 
`expected` is not telling that the value is the current one. It can be an average (ex on optimizers) that can change depending on how you compute things. Here, the value is precisely the current one.

MerlinEgalite:
> Why `expected` and not `current`? `expected` is not telling that the value is the current one. It can be an average (ex on optimizers) that can change depending on how you compute things. Here, the value is precisely the current one.

It's the current one if you accrue the interest in this block. 

But depending on the implementation of the IRM the supply balance evolution might not be monotonic, see [this example](https://github.com/cantinasec/review-morpho-blue-1/pull/68#discussion_r1368042527). The expected balance queried at block 1 could be lower than the one at a later block. As long as it's not updated the supply balance of someone is not set in stone. So I think this is better to keep the "expected" keyword.

julien-devatom:
> ne at a later block. As long as it's not updated the supply balance of someone is not set in stone. So I think this is better to keep the "expected" keyword.

current = in the current block. This is always true that the balance is different in different blocks, whatever the irm. 

However, with bad debt realization, it's true to say that the balance can be different in the current block.

MerlinEgalite:
> current = in the current block. This is always true that the balance is different in different blocks, whatever the irm.

Note that's it's not always true if there's no borrow.

> However, with bad debt realization, it's true to say that the balance can be different in the current block.

So current is not accurate in this case, right?

test/forge/libraries/periphery/MorphoBalancesLibTest.sol
```diff
@@ -73,31 +73,31 @@ contract MorphoBalancesLibTest is BaseTest {
     {
         _generatePendingInterest(amountSupplied, amountBorrowed, timeElapsed, fee);
 
-        uint256 expectedSupplyBalance = morpho.expectedSupplyBalance(marketParams, address(this));
+        uint256 expectedSupplyAssets = morpho.expectedSupplyAssets(marketParams, address(this));
```
QGarchery, line 76:
```suggestion
        uint256 expectedSupplyBalance = morpho.expectedSupplyAssets(marketParams, address(this));
```
so that we keep the same naming for the expected value and the actual value. I don't think that having balance is a problem here, the naming "balance" is used liberally in a lot of places and it's fine

test/forge/libraries/periphery/MorphoBalancesLibTest.sol
```diff
@@ -73,31 +73,31 @@ contract MorphoBalancesLibTest is BaseTest {
...
+        assertEq(expectedSupplyAssets, actualSupplyBalance);
     }
 
     function testExpectedBorrowBalance(uint256 amountSupplied, uint256 amountBorrowed, uint256 timeElapsed, uint256 fee)
         internal
     {
         _generatePendingInterest(amountSupplied, amountBorrowed, timeElapsed, fee);
 
-        uint256 expectedBorrowBalance = morpho.expectedBorrowBalance(marketParams, address(this));
+        uint256 expectedBorrowAssets = morpho.expectedBorrowAssets(marketParams, address(this));
```
QGarchery, line 92:
```suggestion
        uint256 expectedBorrowBalance = morpho.expectedBorrowAssets(marketParams, address(this));
```
same reasoning

test/forge/libraries/periphery/MorphoBalancesLibTest.sol
QGarchery: > Perhaps we should update the test name instead?

It wouldn't help with the duality expected/actual. In any case this is not a blocker for me

src/libraries/periphery/MorphoBalancesLib.sol
```diff
@@ -105,8 +105,8 @@ library MorphoBalancesLib {
         return supplyShares.toAssetsDown(totalSupplyAssets, totalSupplyShares);
     }
 
-    /// @notice Returns the expected borrow balance of a user on a market after having accrued interest.
-    function expectedBorrowBalance(IMorpho morpho, MarketParams memory marketParams, address user)
+    /// @notice Returns the expected borrow assets balance of a `user` on a market after having accrued interest.
```
Rubilmax, line 108:
```suggestion
    /// @notice Returns the expected borrow assets balance of `user` on a market after having accrued interest.
```

src/libraries/periphery/MorphoBalancesLib.sol
```diff
@@ -91,9 +91,9 @@ library MorphoBalancesLib {
         (, totalSupplyShares,,) = expectedMarketBalances(morpho, marketParams);
     }
 
-    /// @notice Returns the expected supply balance of a user on a market after having accrued interest.
+    /// @notice Returns the expected supply assets balance of a `user` on a market after having accrued interest.
```
Rubilmax, line 94:
```suggestion
    /// @notice Returns the expected supply assets balance of `user` on a market after having accrued interest.
```

Rubilmax (approved):

Rubilmax (approved):

MathisGD:
`updated` can make everyone agree ?

QGarchery (approved):

julien-devatom (approved):
@tomrpl To consider for your snippets

Jean-Grimal (approved):

PR #552, opened 2023-10-18
Add name in EIP712 domain

MerlinEgalite:
Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/38

Why we did not include it:
https://github.com/morpho-org/morpho-blue/pull/425

To be discussed.

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -68,7 +68,7 @@ contract Morpho is IMorpho {
         require(newOwner != address(0), ErrorsLib.ZERO_ADDRESS);
 
         owner = newOwner;
-        DOMAIN_SEPARATOR = keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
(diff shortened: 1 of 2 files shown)


Rubilmax (approved):

Jean-Grimal (approved):

MathisGD (changes requested):
I'm still against, I think that this is more dangerous than useful

Rubilmax:
> I'm still against, I think that this is more dangerous than useful

How is it dangerous? The chain id and the verifying contract's address is hashed into the domain, so it's only a matter of UI?

Rubilmax:
> https://github.com/morpho-org/morpho-blue/pull/425#issue-1863332811

Ah right, fair point!

MerlinEgalite:
Stermi POV: https://github.com/cantinasec/review-morpho-blue-1/issues/38#issuecomment-1770483067

PR #553, opened 2023-10-18
Acknowledge EIP-3156 non compliance

MerlinEgalite:
- Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/45

### DIFF
--- a/src/interfaces/IMorpho.sol
+++ b/src/interfaces/IMorpho.sol
@@ -277,6 +277,11 @@ interface IMorpho {
     /// @notice Executes a flash loan.
     /// @dev Flash loans have access to the whole balance of the contract (the liquidity and deposited collateral of all
     /// markets combined, plus donations).
+    /// @dev Warning: Not ERC-3156 compliant for the following reasons:
+    /// - It would add an unnecessary `flashFee` function since there's no fee on Morpho Blue.
+    /// - `maxFlashLoan` is the token balance of the contract.
+    /// - The receiver of the `assets` is the caller as there's no `receiver` argument in our implementation.
+    /// - The ERC is not widespread.
     /// @param token The token to flash loan.
     /// @param assets The amount of assets to flash loan.
     /// @param data Arbitrary data to pass to the `onMorphoFlashLoan` callback.
--- a/src/interfaces/IMorpho.sol
+++ b/src/interfaces/IMorpho.sol
@@ -277,11 +277,10 @@ interface IMorpho {
     /// @notice Executes a flash loan.
     /// @dev Flash loans have access to the whole balance of the contract (the liquidity and deposited collateral of all
     /// markets combined, plus donations).
-    /// @dev Warning: Not ERC-3156 compliant for the following reasons:
-    /// - It would add an unnecessary `flashFee` function since there's no fee on Morpho Blue.
-    /// - `maxFlashLoan` is the token balance of the contract.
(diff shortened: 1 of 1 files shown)


src/interfaces/IMorpho.sol
```diff
@@ -277,6 +277,11 @@ interface IMorpho {
     /// @notice Executes a flash loan.
     /// @dev Flash loans have access to the whole balance of the contract (the liquidity and deposited collateral of all
     /// markets combined, plus donations).
+    /// @dev Warning: Not ERC-3156 compliant for the following reasons:
+    /// - It would add an unnecessary `flashFee` function since there's no fee on Morpho Blue.
+    /// - `maxFlashLoan` is the token balance of the contract.
+    /// - The receiver of the `assets` is the caller as there's no `receiver` argument in our implementation.
+    /// - The ERC is not widespread.
```
Rubilmax, line 284:
```suggestion
    /// @dev Warning: Not ERC-3156 compliant but compatibility is easily reached:
    /// - `flashFee` is zero.
    /// - `maxFlashLoan` is the token's balance of this contract.
    /// - The receiver of `assets` is the caller.
```

Rubilmax (changes requested):
I don't think we should motivate why we don't like the EIP but rather explain that it's simply not compatible, and perhaps highlight how to use Blue's implementation to someone who's used to ERC3156

MerlinEgalite:
I agree but what stermi was asking is to give the reasons why we are not implementing it though. In doing what you propose, we're not answering their concern if I understand it well but perhaps it's better as we don't open the door to any debate.

Rubilmax:
> I agree but what stermi was asking is to give the reasons why we are not implementing it though. In doing what you propose, we're not answering their concern if I understand it well but perhaps it's better as we don't open the door to any debate.

Indeed, StErMi asked for this but I don't like it and suggest another way

src/interfaces/IMorpho.sol
MathisGD, line 284: hot take we don't care about the reasons 
```suggestion
    /// @dev Warning: Not ERC-3156 compliant.
```

Rubilmax: I think we do care about the differences if we want to simplify the integrator's work
And it doesn't cost much to include these: the standard is not expected to change nor the flash loan mechanism on Morpho Blue, so no maintenance cost is expected

MathisGD:
```suggestion
    /// @dev Warning: Not ERC-3156 compliant for the following reasons:
    /// - Morpho Blue doesn't expose the `flashFee` function.
    /// - `maxFlashLoan` is not exposed, but it would be the token balance of the contract.
    /// - The receiver of the `assets` is always the caller.
```

MathisGD: in favor of Romain's suggestion

Rubilmax (approved):

Jean-Grimal (approved):

MathisGD (approved):

MerlinEgalite:
@QGarchery can you have a look please?

QGarchery (approved):

QGarchery:
> I think we do care about the differences if we want to simplify the integrator's work
And it doesn't cost much to include these: the standard is not expected to change nor the flash loan mechanism on Morpho Blue, so no maintenance cost is expected

It kind of applies to just implementing the standard too. If it was widely accepted it would be an easy choice. It's not, but at the same time it's not obvious if it won't be in the future

PR #554, opened 2023-10-18
Clarify fee recipient documentation

MerlinEgalite:
- Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/44
- Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/43

### DIFF
--- a/src/interfaces/IMorpho.sol
+++ b/src/interfaces/IMorpho.sol
@@ -125,7 +125,8 @@ interface IMorpho {
 
     /// @notice Sets `newFeeRecipient` as `feeRecipient` of the fee.
     /// @dev Warning: If the fee recipient is set to the zero address, fees will accrue there.
-    /// @dev Modifying the fee recipient will allow the new recipient to claim any pending fees not yet accrued. To ensure that the current recipient receives all due fees, accrue interest manually prior to making any changes.
+    /// @dev Modifying the fee recipient will allow the new recipient to claim any pending fees not yet accrued. To
+    /// ensure that the current recipient receives all due fees, accrue interest manually prior to making any changes.
     function setFeeRecipient(address newFeeRecipient) external;
 
     /// @notice Creates the market `marketParams`.
--- a/src/interfaces/IMorpho.sol
+++ b/src/interfaces/IMorpho.sol
@@ -106,7 +106,7 @@ interface IMorpho {
         view
         returns (address loanToken, address collateralToken, address oracle, address irm, uint256 lltv);
 
-    /// @notice Sets `newOwner` as owner of the contract.
+    /// @notice Sets `newOwner` as `owner` of the contract.
     /// @dev Warning: No two-step transfer ownership.
     /// @dev Warning: The owner can be set to the zero address.
     function setOwner(address newOwner) external;
@@ -123,10 +123,13 @@ interface IMorpho {
     /// @dev Warning: The recipient can be the zero address.
     function setFee(MarketParams memory marketParams, uint256 newFee) external;
 
-    /// @notice Sets `newFeeRecipient` as recipient of the fee.
-    /// @dev Warning: The fee recipient can be set to the zero address.
-    /// @dev Warning: The fee to be accrued on each market won't belong to the old fee recipient after calling this
-    /// function.
+    /// @notice Sets `newFeeRecipient` as `feeRecipient` of the fee.
+    /// @dev Warning: The fee recipient can be set to the zero address. Setting the fee recipient to this address
+    /// would update the zero address position on each market having a fee, thus, burning the fee. It's the owner
+    /// responsibility to make sure no market has a fee before setting the fee recipient to the zero address.
+    /// @dev Warning: When setting a new fee recipient, interest is not accrued on each market having a fee.
+    /// It's the responsibility of the fee recipient to accrue interest on those markets to get the full amount of fees.
+    /// The remainder will go to the new fee recipient.
     function setFeeRecipient(address newFeeRecipient) external;
 
     /// @notice Creates the market `marketParams`.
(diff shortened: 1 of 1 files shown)


Rubilmax (approved):

src/interfaces/IMorpho.sol
```diff
@@ -123,10 +123,13 @@ interface IMorpho {
...
     function setFee(MarketParams memory marketParams, uint256 newFee) external;
 
-    /// @notice Sets `newFeeRecipient` as recipient of the fee.
-    /// @dev Warning: The fee recipient can be set to the zero address.
-    /// @dev Warning: The fee to be accrued on each market won't belong to the old fee recipient after calling this
-    /// function.
+    /// @notice Sets `newFeeRecipient` as `feeRecipient` of the fee.
+    /// @dev Warning: The fee recipient can be set to the zero address. Setting the fee recipient to this address
+    /// would update the zero address position on each market having a fee, thus, burning the fee. It's the owner
+    /// responsibility to make sure no market has a fee before setting the fee recipient to the zero address.
```
MathisGD, line 129: third part feels obvious once we explained this, this is enough:
```suggestion
    /// @dev Warning: If the fee recipient is set to the zero address, fees will accrue there.
```

src/interfaces/IMorpho.sol
```diff
@@ -123,10 +123,13 @@ interface IMorpho {
...
+    /// @dev Warning: When setting a new fee recipient, interest is not accrued on each market having a fee.
```
Jean-Grimal, line 130:
```suggestion
    /// @dev Warning: When setting a new fee recipient, interest is not instantly accrued on each market having a fee.
```
I feel like it makes it clearer

MathisGD (approved):

src/interfaces/IMorpho.sol
```diff
@@ -123,10 +123,10 @@ interface IMorpho {
     /// @dev Warning: The recipient can be the zero address.
     function setFee(MarketParams memory marketParams, uint256 newFee) external;
 
-    /// @notice Sets `newFeeRecipient` as recipient of the fee.
-    /// @dev Warning: The fee recipient can be set to the zero address.
-    /// @dev Warning: The fee to be accrued on each market won't belong to the old fee recipient after calling this
-    /// function.
+    /// @notice Sets `newFeeRecipient` as `feeRecipient` of the fee.
+    /// @dev Warning: If the fee recipient is set to the zero address, fees will accrue there.
```
Jean-Grimal, line 127:
```suggestion
    /// @dev Warning: If the fee recipient is set to the zero address, fees will continue to accrue but will be lost.
```

Jean-Grimal (changes requested):

src/interfaces/IMorpho.sol
MathisGD:
```suggestion
    /// @dev Warning: If the fee recipient is set to the zero address, fees will accrue there and will be lost.
```

Rubilmax (approved):

Jean-Grimal (approved):

MathisGD (approved):

src/interfaces/IMorpho.sol
```diff
@@ -123,10 +123,10 @@ interface IMorpho {
     /// @dev Warning: The recipient can be the zero address.
     function setFee(MarketParams memory marketParams, uint256 newFee) external;
 
-    /// @notice Sets `newFeeRecipient` as recipient of the fee.
-    /// @dev Warning: The fee recipient can be set to the zero address.
-    /// @dev Warning: The fee to be accrued on each market won't belong to the old fee recipient after calling this
-    /// function.
+    /// @notice Sets `newFeeRecipient` as `feeRecipient` of the fee.
```
QGarchery, line 126: same

src/interfaces/IMorpho.sol
```diff
@@ -123,10 +123,10 @@ interface IMorpho {
...
+    /// @dev Warning: If the fee recipient is set to the zero address, fees will accrue there and will be lost.
+    /// @dev Modifying the fee recipient will allow the new recipient to claim any pending fees not yet accrued. To
+    /// ensure that the current recipient receives all due fees, accrue interest manually prior to making any changes.
```
QGarchery, line 129: Mentioning the use case of accruing interest really makes me wonder if we should make the `accrueInterest` function public

QGarchery (changes requested):

src/interfaces/IMorpho.sol
```diff
@@ -106,7 +106,7 @@ interface IMorpho {
         view
         returns (address loanToken, address collateralToken, address oracle, address irm, uint256 lltv);
 
-    /// @notice Sets `newOwner` as owner of the contract.
+    /// @notice Sets `newOwner` as `owner` of the contract.
```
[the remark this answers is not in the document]
MerlinEgalite: Does it work with global variable as well? But in this case it would not work because it should be `owner()` perhaps

src/interfaces/IMorpho.sol
MerlinEgalite: I tend to think we should, it would allow integrators to have more certainty on their actions I think

src/interfaces/IMorpho.sol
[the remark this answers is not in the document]
Rubilmax: Source: https://docs.soliditylang.org/en/v0.8.17/natspec-format.html#dynamic-expressions

But it doesn't state that it should be limited to local variables so I'm fine with using this syntax to highlight global or local variables and don't have a strict convention on its usage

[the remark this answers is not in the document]
QGarchery: Ok let's do this

src/interfaces/IMorpho.sol
QGarchery: Let's discuss that separately, here is the [related thread](https://morpholabs.slack.com/archives/C02N7CZ088N/p1698164347788439)

QGarchery (approved):

PR #556, opened 2023-10-22
perf: minor gas improvements

antoncoding:
## Summary
Minor gas savings by removing unnecessary overflow / underflow checks.

 ## Details
This is inspired by [OpenZepplin's ERC20 implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5ae630684a0f57de400ef69499addab4c32ac8fb/contracts/token/ERC20/ERC20.sol#L259): when dealing with updating balances & total supply:
* minting: if total supply doesn't overflow, individual's balance cannot overflow (total supply = sum of all balances)
* burning: if individual's balance doesn't underflow, total supply cannot underflow  

The very same concept can be applied here: while updating user shares, this is the same as above: when we update each account's shares, like `position[id][onBehalf].supplyShares` and `market[id].totalSupplyShares`, we only need 1 overall check: when checking += , only check the aggregated totalSupply or totalBorrow can overflow, when checking -=, only check if each users' position underflow

### P.S.
The improvement is minor, according to snapshot maybe scale of < 100 - 200. it can probably be extended a bit more on more storage updates, like across share and assets, happy to help exploring further. But feel free to close if this is not something you want to focus on.

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -170,9 +170,12 @@ contract Morpho is IMorpho {
         if (assets > 0) shares = assets.toSharesDown(market[id].totalSupplyAssets, market[id].totalSupplyShares);
         else assets = shares.toAssetsUp(market[id].totalSupplyAssets, market[id].totalSupplyShares);
 
-        position[id][onBehalf].supplyShares += shares;
         market[id].totalSupplyShares += shares.toUint128();
         market[id].totalSupplyAssets += assets.toUint128();
+
+        unchecked {
+            position[id][onBehalf].supplyShares += shares;
+        }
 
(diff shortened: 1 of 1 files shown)


Rubilmax:
We've considered doing this in the past but here is our reasoning:

According to our hardhat gas tests, the savings are in the order of magnitude of 100 gas **maximum**, while an interaction on Morpho Blue can range from 70k to 100k gas ; meaning such an "optimization" would earn more or less ~0.1% of the total tx gas cost
While such a change impacts readability and complexifies the reasoning over the code (we can't trust the compiler anymore to revert in case of overflow/underflow)

For these reasons, we won't be making such changes

PR #557, opened 2023-10-23
`UtilsLib.zeroFloorSub` is not utilised consistently

MerlinEgalite:
- Fixes #https://github.com/cantinasec/review-morpho-blue-1/issues/58

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -377,16 +377,14 @@ contract Morpho is IMorpho {
         uint256 badDebtShares;
         if (position[id][borrower].collateral == 0) {
             badDebtShares = position[id][borrower].borrowShares;
-            position[id][borrower].borrowShares = 0;
-
-            uint128 totalBorrowAssets = market[id].totalBorrowAssets;
             uint128 badDebt = UtilsLib.min(
-                totalBorrowAssets, badDebtShares.toAssetsUp(totalBorrowAssets, market[id].totalBorrowShares)
+                market[id].totalBorrowAssets,
+                badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares)
             ).toUint128();
-
             market[id].totalBorrowAssets -= badDebt;
             market[id].totalSupplyAssets -= badDebt;
             market[id].totalBorrowShares -= badDebtShares.toUint128();
+            position[id][borrower].borrowShares = 0;
         }
 
         IERC20(marketParams.collateralToken).safeTransfer(msg.sender, seizedAssets);
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -377,7 +377,7 @@ contract Morpho is IMorpho {
         uint256 badDebtShares;
         if (position[id][borrower].collateral == 0) {
             badDebtShares = position[id][borrower].borrowShares;
-            // Safe "unchecked" cast since badDebt <= totalBorrowAssets <= type(uint128).max
+            // Safe "unchecked" cast since the value is smaller than totalBorrowAssets.
             uint128 badDebt = uint128(
                 UtilsLib.min(
                     market[id].totalBorrowAssets,
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -380,12 +380,12 @@ contract Morpho is IMorpho {
             position[id][borrower].borrowShares = 0;
 
             uint128 totalBorrowAssets = market[id].totalBorrowAssets;
-            uint256 badDebt = badDebtShares.toAssetsUp(totalBorrowAssets, market[id].totalBorrowShares);
-            uint128 newTotalBorrowAssets = UtilsLib.zeroFloorSub(totalBorrowAssets, badDebt).toUint128();
-            uint128 newBadDebt = totalBorrowAssets - newTotalBorrowAssets;
+            uint128 badDebt = UtilsLib.min(
+                totalBorrowAssets, badDebtShares.toAssetsUp(totalBorrowAssets, market[id].totalBorrowShares)
+            ).toUint128();
 
-            market[id].totalBorrowAssets = newTotalBorrowAssets;
-            market[id].totalSupplyAssets -= newBadDebt;
+            market[id].totalBorrowAssets -= badDebt;
+            market[id].totalSupplyAssets -= badDebt;
             market[id].totalBorrowShares -= badDebtShares.toUint128();
         }
 
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -377,16 +377,14 @@ contract Morpho is IMorpho {
         uint256 badDebtShares;
         if (position[id][borrower].collateral == 0) {
             badDebtShares = position[id][borrower].borrowShares;
-            // Safe "unchecked" cast since the value is smaller than totalBorrowAssets.
-            uint128 badDebt = uint128(
-                UtilsLib.min(
-                    market[id].totalBorrowAssets,
-                    badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares)
-                )
+            uint256 badDebt = UtilsLib.min(
+                market[id].totalBorrowAssets,
+                badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares)
             );
-            market[id].totalBorrowAssets -= badDebt;
-            market[id].totalSupplyAssets -= badDebt;
-            market[id].totalBorrowShares -= uint128(badDebtShares); // Safe "unchecked" cast.
+
+            market[id].totalBorrowAssets -= badDebt.toUint128();
+            market[id].totalSupplyAssets -= badDebt.toUint128();
+            market[id].totalBorrowShares -= badDebtShares.toUint128();
             position[id][borrower].borrowShares = 0;
         }
 
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -377,13 +377,16 @@ contract Morpho is IMorpho {
         uint256 badDebtShares;
         if (position[id][borrower].collateral == 0) {
             badDebtShares = position[id][borrower].borrowShares;
-            uint128 badDebt = UtilsLib.min(
-                market[id].totalBorrowAssets,
-                badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares)
-            ).toUint128();
+            // Safe "unchecked" cast since badDebt <= totalBorrowAssets <= type(uint128).max
+            uint128 badDebt = uint128(
+                UtilsLib.min(
+                    market[id].totalBorrowAssets,
+                    badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares)
+                )
+            );
             market[id].totalBorrowAssets -= badDebt;
             market[id].totalSupplyAssets -= badDebt;
-            market[id].totalBorrowShares -= badDebtShares.toUint128();
+            market[id].totalBorrowShares -= uint128(badDebtShares); // Safe "unchecked" cast.
             position[id][borrower].borrowShares = 0;
         }
 
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -377,11 +377,16 @@ contract Morpho is IMorpho {
         uint256 badDebtShares;
         if (position[id][borrower].collateral == 0) {
             badDebtShares = position[id][borrower].borrowShares;
-            uint256 badDebt = badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares);
-            market[id].totalSupplyAssets -= badDebt.toUint128();
-            market[id].totalBorrowAssets -= badDebt.toUint128();
-            market[id].totalBorrowShares -= badDebtShares.toUint128();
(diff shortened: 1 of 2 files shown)


src/Morpho.sol
```diff
@@ -377,11 +377,16 @@ contract Morpho is IMorpho {
         uint256 badDebtShares;
         if (position[id][borrower].collateral == 0) {
             badDebtShares = position[id][borrower].borrowShares;
-            uint256 badDebt = badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares);
-            market[id].totalSupplyAssets -= badDebt.toUint128();
-            market[id].totalBorrowAssets -= badDebt.toUint128();
-            market[id].totalBorrowShares -= badDebtShares.toUint128();
             position[id][borrower].borrowShares = 0;
```
MerlinEgalite, line 380: This change is mandatory to empty the stack and avoid enforcing to use via ir to compile the contract

Rubilmax: to free up 1 stack slot formerly occupied by `borrower`*

right?

src/Morpho.sol
```diff
@@ -377,11 +377,16 @@ contract Morpho is IMorpho {
...
+
+            uint128 totalBorrowAssets = market[id].totalBorrowAssets;
+            uint256 badDebt = badDebtShares.toAssetsUp(totalBorrowAssets, market[id].totalBorrowShares);
+            uint128 newTotalBorrowAssets = UtilsLib.zeroFloorSub(totalBorrowAssets, badDebt).toUint128();
+            uint128 newBadDebt = totalBorrowAssets - newTotalBorrowAssets;
```
Rubilmax, line 385: wouldn't it be equivalent but easier to have `badDebt = UtilsLib.min(totalBorrowAssets, badDebtShares.toAssetsUp(totalBorrowAssets, market[id].totalBorrowShares))`?

MerlinEgalite: I'll try

src/Morpho.sol
MerlinEgalite: It's not needed anymore with your suggestion but I would keep it, wdyt?

src/Morpho.sol
```diff
@@ -377,11 +377,16 @@ contract Morpho is IMorpho {
         uint256 badDebtShares;
         if (position[id][borrower].collateral == 0) {
             badDebtShares = position[id][borrower].borrowShares;
-            uint256 badDebt = badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares);
-            market[id].totalSupplyAssets -= badDebt.toUint128();
-            market[id].totalBorrowAssets -= badDebt.toUint128();
-            market[id].totalBorrowShares -= badDebtShares.toUint128();
             position[id][borrower].borrowShares = 0;
+
+            uint128 totalBorrowAssets = market[id].totalBorrowAssets;
```
Rubilmax, line 382: Is this stack assignment necessary? It goes against minimalist conventions we took for Blue (of relying on the compiler for optimizations)

src/Morpho.sol
Rubilmax: I'm ambivalent,

Jean-Grimal (approved):

src/Morpho.sol
MerlinEgalite: Should I @MathisGD ?

src/Morpho.sol
```diff
@@ -377,11 +377,16 @@ contract Morpho is IMorpho {
...
+            uint128 badDebt = UtilsLib.min(
+                totalBorrowAssets, badDebtShares.toAssetsUp(totalBorrowAssets, market[id].totalBorrowShares)
+            ).toUint128();
```
QGarchery, line 385: Here is a failing test for this change:
```solidity

    function testBadDebtDoesNotOverflow() public {
        {
            uint256 collateralAmount = 10 ether;
            uint256 loanAmount = 1 ether;
            _supply(loanAmount);

            collateralToken.setBalance(BORROWER, collateralAmount);
            vm.startPrank(BORROWER);
            morpho.supplyCollateral(marketParams, collateralAmount, BORROWER, hex"");
            morpho.borrow(marketParams, loanAmount, 0, BORROWER, BORROWER);
            // Trick to inflate shares, so that the computed bad debt is greater than the total debt of the market.
            morpho.borrow(marketParams, 0, 1, BORROWER, BORROWER);
            vm.stopPrank();

            oracle.setPrice(1e36 / 100);

            loanToken.setBalance(LIQUIDATOR, loanAmount);
            vm.prank(LIQUIDATOR);
            morpho.liquidate(marketParams, BORROWER, collateralAmount, 0, hex"");
        }
    }
```

QGarchery (changes requested):
Can we rename this PR please for future reference ?

Looks good to me, only a few minor changes

src/Morpho.sol
MerlinEgalite: To me it's more readable this way

src/Morpho.sol
Rubilmax: So what should we do?

Rubilmax: Oh I get it, we should not cast `badDebt` to `uint128` is that it?

Rubilmax: Btw casting it to `uint128` could be unchecked because `badDebt <= totalBorrowAssets <= type(uint128).max`

QGarchery: > So what should we do?

Let's add this test !

MerlinEgalite: It's also valid for `badDebtShares` btw!

Jean-Grimal (approved):
Nice

test/forge/integration/LiquidateIntegrationTest.sol
```diff
@@ -296,4 +296,24 @@ contract LiquidateIntegrationTest is BaseTest {
             morpho.totalSupplyAssets(id), params.totalSupplyBeforeLiquidation - params.expectedBadDebt, "total supply"
         );
     }
+
+    function testBadDebtDoesNotOverflow() public {
```
Rubilmax, line 300:
```suggestion
    function testBadDebtOverTotalBorrowAssets() public {
```

Rubilmax (approved):

src/Morpho.sol
```diff
@@ -377,10 +377,16 @@ contract Morpho is IMorpho {
...
+            // Safe "unchecked" cast since badDebt <= totalBorrowAssets <= type(uint128).max
+            uint128 badDebt = uint128(
+                UtilsLib.min(
+                    market[id].totalBorrowAssets,
+                    badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares)
+                )
+            );
+            market[id].totalBorrowAssets -= badDebt;
+            market[id].totalSupplyAssets -= badDebt;
+            market[id].totalBorrowShares -= uint128(badDebtShares); // Safe "unchecked" cast.
```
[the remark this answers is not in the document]
Rubilmax: I think there's another reason this was not done before. Do you have one in mind @MathisGD @QGarchery ?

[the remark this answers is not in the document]
MerlinEgalite: I don't remember tbh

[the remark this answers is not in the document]
QGarchery: I think it was just for consistency with other cases where we did the casting. Since this is gone anyway, I think we can do that here yes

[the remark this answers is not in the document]
QGarchery: The `badDebtShares` is going to be emitted as a `uint256` though, so I'm not sure if it saves gas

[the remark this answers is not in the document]
Rubilmax: There's also another unnecessary usage of `toUint128` at L372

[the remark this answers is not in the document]
MerlinEgalite: Agree should I bundle it in this PR?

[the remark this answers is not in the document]
Rubilmax: I'm not against but if we want to be clean it should be done in a dedicated PR

Jean-Grimal (dismissed):

src/Morpho.sol
```diff
@@ -377,10 +377,16 @@ contract Morpho is IMorpho {
         uint256 badDebtShares;
         if (position[id][borrower].collateral == 0) {
             badDebtShares = position[id][borrower].borrowShares;
-            uint256 badDebt = badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares);
-            market[id].totalSupplyAssets -= badDebt.toUint128();
-            market[id].totalBorrowAssets -= badDebt.toUint128();
-            market[id].totalBorrowShares -= badDebtShares.toUint128();
+            // Safe "unchecked" cast since badDebt <= totalBorrowAssets <= type(uint128).max
```
QGarchery, line 380:
```suggestion
            // Safe "unchecked" cast since the value is smaller than totalBorrowAssets
```
because badDebt is the result of the cast. But maybe it's clearer without this suggestion, even if it's a little incorrect

QGarchery (approved):

src/Morpho.sol
MerlinEgalite, line 380:
```suggestion
            // Safe "unchecked" cast since the value is smaller than totalBorrowAssets.
```
```suggestion
            // Safe "unchecked" cast since the value is smaller than totalBorrowAssets.
```

MerlinEgalite: adding the point

Rubilmax: I'm fine with both

Rubilmax (dismissed):

QGarchery (dismissed):

src/Morpho.sol
```diff
@@ -377,10 +377,16 @@ contract Morpho is IMorpho {
...
-            market[id].totalSupplyAssets -= badDebt.toUint128();
-            market[id].totalBorrowAssets -= badDebt.toUint128();
-            market[id].totalBorrowShares -= badDebtShares.toUint128();
+            // Safe "unchecked" cast since the value is smaller than totalBorrowAssets.
+            uint128 badDebt = uint128(
+                UtilsLib.min(
+                    market[id].totalBorrowAssets,
+                    badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares)
+                )
+            );
```
MathisGD, line 386: I'm more in favor of casting, because we do it everywhere already

MerlinEgalite: You mean safe casting?

MerlinEgalite: If yes, it's not really true. There are 4 other occurrences where we don't safe cast

Jean-Grimal (approved):

src/Morpho.sol
MathisGD: ah yes nice catch. I find it consistent to have this one "checked", but no veto

Rubilmax: Arbitrarily consistent then (I disagree for the same reason as Merlin)

MathisGD: ok then let's just check everything and use only uint256 variables

Rubilmax: I assume you suggest this for consistency
What's the benefit of being consistent in this case?
I'd rather go for the status quo

MathisGD: >I'd rather go for the status quo to minimize changes

agree but we disagree on what is the status quo here (in this PR to me we diverge from the status quo)

Rubilmax: Indeed, I meant I am against changing anything elsewhere (which I thought was what you meant with  "let's just check everything and use only uint256 variables")

And because we are modifying the line in this PR we have the free choice to either safe cast it or not
There seems to be no consistency in the rest of the code ; though we could argue that we TRIED to not safe cast when unnecessary - so I'd argue in favor of not safe casting it here

MathisGD: @Rubilmax see #573 wdyt

MerlinEgalite: Let's merge this PR as it is and that's it, no?

MerlinEgalite:
I think we should merge as is

src/Morpho.sol
MathisGD: no let's just have a uint256 as it was, and safe cast

MerlinEgalite: 🥲

src/Morpho.sol
```diff
@@ -377,10 +377,16 @@ contract Morpho is IMorpho {
...
+            market[id].totalBorrowAssets -= badDebt;
+            market[id].totalSupplyAssets -= badDebt;
+            market[id].totalBorrowShares -= uint128(badDebtShares); // Safe "unchecked" cast.
```
MathisGD, line 389:
```suggestion
            uint256 badDebt = UtilsLib.min(
                market[id].totalBorrowAssets,
                badDebtShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares)
            );

            market[id].totalBorrowAssets -= badDebt.toUint128();
            market[id].totalSupplyAssets -= badDebt.toUint128();
            market[id].totalBorrowShares -= badDebtShares.toUint128();
```

MathisGD (approved):

MerlinEgalite (commented):
LGTM

QGarchery (approved):

Rubilmax (approved):

PR #559, opened 2023-10-24
Add some missing comment

MerlinEgalite:
- Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/15
- Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/62

I did not include every suggestions since some are unnecessary to me


src/interfaces/IMorpho.sol
```diff
@@ -139,6 +144,8 @@ interface IMorpho {
     /// - The token balance of the sender (resp. receiver) should decrease (resp. increase) by exactly the given amount
     /// on `transfer` and `transferFrom`. In particular, tokens with fees on transfer are not supported.
     /// - The IRM should not re-enter Morpho.
+    /// - The oracle should return a price with the correct scaling.
+    /// - The oracle is trusted to handle stale prices and price manipulations if necessary.
```
Rubilmax, line 148: Is this last assumption correct actually? I don't think we make such assumption on the oracle

src/libraries/MarketParamsLib.sol
```diff
@@ -12,7 +12,7 @@ library MarketParamsLib {
     /// @dev The length is 5 * 32 because `MarketParams` has 5 variables of 32 bytes each.
     uint256 internal constant MARKET_PARAMS_BYTES_LENGTH = 5 * 32;
 
-    /// @notice Returns the id of the market `marketParams`.
+    /// @notice Returns the id (`marketParamsId`) of the market `marketParams`.
```
Rubilmax, line 15: I actually think it's clearer without this ^^

src/libraries/periphery/MorphoLib.sol
```diff
@@ -8,6 +8,8 @@ import {MorphoStorageLib} from "./MorphoStorageLib.sol";
 /// @author Morpho Labs
 /// @custom:contact security@morpho.org
 /// @notice Helper library to access Morpho storage variables.
+/// @dev Warning: Supply and borrow getters may returned outdated values that do not include the accrued interest and
```
Rubilmax, line 11:
```suggestion
/// @dev Warning: Supply and borrow getters may returned outdated values that do not include accrued interest and
```

src/libraries/MarketParamsLib.sol
MerlinEgalite: will remove then

src/interfaces/IMorpho.sol
MerlinEgalite: The comment above is the following:

```solidity
    /// @dev Here is the list of assumptions on the market's dependencies (tokens, IRM and oracle) that guarantees
    /// Morpho behaves as expected:
```

How to we say that the oracle must behave well as well?

Jean-Grimal (dismissed):

src/Morpho.sol
```diff
@@ -194,8 +194,8 @@ contract Morpho is IMorpho {
         Id id = marketParams.id();
         require(market[id].lastUpdate != 0, ErrorsLib.MARKET_NOT_CREATED);
         require(UtilsLib.exactlyOneZero(assets, shares), ErrorsLib.INCONSISTENT_INPUT);
-        // No need to verify that onBehalf != address(0) thanks to the authorization check.
         require(receiver != address(0), ErrorsLib.ZERO_ADDRESS);
+        // No need to verify that onBehalf != address(0) thanks to the authorization check.
```
MathisGD, line 198: there is a good reason why it's here: it's the position of the onBehalf != address(0) check in other functions (where it is checked). And I'm not sure that putting it here improves anything.

src/interfaces/IMorpho.sol
```diff
@@ -73,6 +73,11 @@ interface IMorpho {
         returns (uint256 supplyShares, uint128 borrowShares, uint128 collateral);
 
     /// @notice The state of the market corresponding to `id`.
+    /// @dev Warning: `totalSupplyAssets` does not contain the accrued interest since the last interest accrual.
+    /// @dev Warning: `totalBorrowAssets` does not contain the accrued interest since the last interest accrual.
+    /// @dev Warning: `totalSupplyShares` does not contain the additional shares accrued by `feeRecipient` since the
+    /// last
+    /// interest accrual.
```
MathisGD, line 80: wrong formatting

src/interfaces/IMorpho.sol
MathisGD: >Is this last assumption correct actually? I don't think we make such assumption on the oracle

yes, this is not needed in order for Morpho to behave as expected

src/libraries/MathLib.sol
```diff
@@ -8,33 +8,33 @@ uint256 constant WAD = 1e18;
...
 /// @notice Library to manage fixed-point arithmetic.
 library MathLib {
-    /// @dev (x * y) / WAD rounded down.
+    /// @dev Returns (`x` * `y`) / `WAD` rounded down.
     function wMulDown(uint256 x, uint256 y) internal pure returns (uint256) {
         return mulDivDown(x, y, WAD);
     }
 
-    /// @dev (x * WAD) / y rounded down.
+    /// @dev Returns (`x` * `WAD`) / `y` rounded down.
```
QGarchery, line 16:
```suggestion
    /// @dev Returns (`x` * WAD) / `y` rounded down.
```
not sure if we should do this actually. So the idea of backticks is to replace the expression dynamically with a concrete value, as explained [here](https://docs.soliditylang.org/en/v0.8.11/natspec-format.html#dynamic-expressions). But after seeing how it is used on this particular line, it feels weird to not do it for every "verbatim" (code-like) expressions

src/Morpho.sol
Rubilmax: I do think it's easier to understand as a reader like so instead of expecting the reader to check the exact position of the corresponding check in other functions (and because the order of these checks doesn't impact the behavior here)

src/interfaces/IMorpho.sol
Rubilmax: Just to be clear, I'm only talking about the line:
`The oracle is trusted to handle stale prices and price manipulations if necessary.`

To me, the assumption:
`The oracle should return a price with the correct scaling.` is required to have Morpho behave as expected, but not the other one (whether price are manipulated or stale are not considered by Morpho: the protocol will just consume the price of the oracle the market is attached to)

MerlinEgalite: But I think the following should be written somewhere:

> the protocol will just consume the price of the oracle the market is attached to, without any assumption on it

src/Morpho.sol
MerlinEgalite: I agree with @Rubilmax I think it's better to apply the change

src/libraries/MathLib.sol
```diff
@@ -8,33 +8,33 @@ uint256 constant WAD = 1e18;
 /// @custom:contact security@morpho.org
 /// @notice Library to manage fixed-point arithmetic.
 library MathLib {
-    /// @dev (x * y) / WAD rounded down.
+    /// @dev Returns (`x` * `y`) / `WAD` rounded down.
```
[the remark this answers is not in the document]
MerlinEgalite: To me we should add `@param` in case there are many parameters or else that those paramaters require explanations.

src/interfaces/IMorpho.sol
```diff
@@ -73,6 +73,11 @@ interface IMorpho {
         returns (uint256 supplyShares, uint128 borrowShares, uint128 collateral);
 
     /// @notice The state of the market corresponding to `id`.
+    /// @dev Warning: `totalSupplyAssets` does not contain the accrued interest since the last interest accrual.
```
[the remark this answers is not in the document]
MerlinEgalite: I think we should yes.

Redundancy helps everyone understand the code and reminds people of the edge cases which is important. I don't think repeating the same thing has downsides.

src/interfaces/IMorpho.sol
MerlinEgalite: Perhaps should we move that comment to the struct? "The oracle is trusted to handle stale prices and price manipulations if necessary."

QGarchery (changes requested):
`lib/solmate` submodule was added by mistake I think

src/libraries/MathLib.sol
QGarchery: The decision taken is to allow to have backticks on global variables, since it's not clear how expression are replaced dynamically anyway

src/interfaces/IMorpho.sol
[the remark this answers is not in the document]
QGarchery: Ok let's duplicate the comments for the `position` function then

src/libraries/MathLib.sol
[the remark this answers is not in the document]
QGarchery: What do you mean by "many" ? More than 1 ? I'm not against your suggestion but I want us to arrive at a conclusion on this.

More specifically, what do you think about the natspecs comments of:
1. the constructor of the Morpho contract
2. the functions in the IIrm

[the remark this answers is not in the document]
MerlinEgalite: Good question, in my mind we can avoid using @param when the description of input parameters and what returns the functions can fit in one line. Else we should use the `@param` statement. 

But "when the description of input parameters and what returns the functions can fit in one line" does not generalize well. 

For the constructor and IIrm we could have a one liner.

Perhaps to avoid debates, we should always use `@param` whatever the situation.

src/interfaces/IIrm.sol
```diff
@@ -7,6 +7,7 @@ import {MarketParams, Market} from "./IMorpho.sol";
 /// @author Morpho Labs
 /// @custom:contact security@morpho.org
 /// @notice Interface that Interest Rate Models (IRMs) used by Morpho must implement.
+/// @dev It is the responsibility of the IRM to handle the compounding of interest if necessary.
```
Rubilmax, line 10:
```suggestion
/// @dev It is the responsibility of the IRM to handle the compounding of interest if desired.
```

src/interfaces/IOracle.sol
```diff
@@ -5,6 +5,7 @@ pragma solidity >=0.5.0;
 /// @author Morpho Labs
 /// @custom:contact security@morpho.org
 /// @notice Interface that oracles used by Morpho must implement.
+/// @dev The oracle is trusted to handle stale prices and price manipulations if necessary.
```
Rubilmax, line 8:
```suggestion
/// @dev The oracle is responsible for handling stale prices and price manipulations if desired.
```

Rubilmax: we can also harmonize "it is the responsibility of" and "... is responsible for" but it's a tiny detail, I honestly don't care

Rubilmax (approved):

src/libraries/periphery/MorphoLib.sol
```diff
@@ -8,6 +8,8 @@ import {MorphoStorageLib} from "./MorphoStorageLib.sol";
 /// @author Morpho Labs
 /// @custom:contact security@morpho.org
 /// @notice Helper library to access Morpho storage variables.
+/// @dev Warning: Supply and borrow getters may returned outdated values that do not include accrued interest and fees
+/// for instance.
```
Rubilmax, line 12:
```suggestion
/// @dev Warning: Supply and borrow getters may return outdated values that do not include accrued interest.
```

I think fees are a cut on the "raw" interest so I don't think they impact how outdated values are here: it's just the accrued interest (which we consider being raw interest minus fees)

src/interfaces/IOracle.sol
```diff
@@ -5,6 +5,7 @@ pragma solidity >=0.5.0;
 /// @author Morpho Labs
 /// @custom:contact security@morpho.org
 /// @notice Interface that oracles used by Morpho must implement.
+/// @dev It is the responsibility of to oracle to handle stale prices and price manipulations if desired.
```
Rubilmax, line 8:
```suggestion
/// @dev It is the responsibility of the oracle to handle stale prices and price manipulations if desired.
```

Rubilmax (approved):

src/Morpho.sol
QGarchery, line 198:
```suggestion
        // No need to verify that onBehalf != address(0) thanks to the following authorization check.
```
optional but now we can do that

src/interfaces/IIrm.sol
```diff
@@ -7,6 +7,7 @@ import {MarketParams, Market} from "./IMorpho.sol";
 /// @author Morpho Labs
 /// @custom:contact security@morpho.org
 /// @notice Interface that Interest Rate Models (IRMs) used by Morpho must implement.
+/// @dev It is the responsibility of the IRM to handle the compounding of interest if desired.
```
QGarchery, line 10: The compounding of interest is already handled by Morpho Blue, see the use of the `wTaylorCompounded` function. It is the compounding of the variation of the rate that is not necessarily taken into account. The issue was that the `speed` of the variation was assumed to be constant in the IRM, so its effect on the rate would not compound

src/libraries/MathLib.sol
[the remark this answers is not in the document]
QGarchery: I like your rule, I think we should apply it to the IIrm and to the constructor

src/interfaces/IIrm.sol
MerlinEgalite: In this case I would remove entirely the comment since it depends even more on how the IRM is built, wdyt?

src/interfaces/IOracle.sol
```diff
@@ -5,6 +5,7 @@ pragma solidity >=0.5.0;
 /// @author Morpho Labs
 /// @custom:contact security@morpho.org
 /// @notice Interface that oracles used by Morpho must implement.
+/// @dev It is the responsibility of the oracle to handle stale prices and price manipulations if desired.
```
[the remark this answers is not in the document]
MerlinEgalite: Because we don't really do it for CL. We can argue that we rely on them to do it though

src/libraries/MathLib.sol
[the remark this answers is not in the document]
MerlinEgalite: Which one? 😄

This one?

> we can avoid using @param when the description of input parameters and what returns the functions can fit in one line. Else we should use the @param statement.

or this one?

> Perhaps to avoid debates, we should always use @param whatever the situation.

src/interfaces/IIrm.sol
QGarchery: Actually yes, now that I see it I think we should remove it

src/interfaces/IOracle.sol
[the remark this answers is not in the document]
QGarchery: Sorry I was only talking about the `if desired` part. I assume that developpers of oracles on Blue would always desire to take into account stale prices and price manipulations (even if it is to just acknowledge them)

src/libraries/MathLib.sol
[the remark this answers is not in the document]
QGarchery: Oops sorry, I meant the first one. Let's not put `@param` comments that are useless if we can avoid it

Jean-Grimal (approved):

src/Morpho.sol
```diff
@@ -229,8 +229,8 @@ contract Morpho is IMorpho {
...
```
QGarchery, line 233:
```suggestion
        // No need to verify that onBehalf != address(0) thanks to the following authorization check.
```

QGarchery: I should have added those as well

Jean-Grimal (dismissed):

src/libraries/periphery/MorphoLib.sol
QGarchery: Actually in the comment we don't mention "raw interests", so it can work for both. Marking this a resolved then

src/libraries/MathLib.sol
```diff
@@ -8,33 +8,33 @@ uint256 constant WAD = 1e18;
 /// @custom:contact security@morpho.org
 /// @notice Library to manage fixed-point arithmetic.
 library MathLib {
-    /// @dev (x * y) / WAD rounded down.
```
QGarchery, line 11: Can you remove unnecessary comments in the constructor and in IIrm please ?

src/interfaces/IMorpho.sol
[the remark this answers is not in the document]
QGarchery: It is not resolved yet right ? We should duplicate the warning comment that is located over the `Position` struct to the `position` function

src/libraries/MathLib.sol
MerlinEgalite: Is it related to the line of your comment or not at all?

QGarchery: Ah no it is not related to this line. I did not know where to put it, I'll explain that it is not related next time

src/Morpho.sol
```diff
@@ -194,8 +194,8 @@ contract Morpho is IMorpho {
         Id id = marketParams.id();
         require(market[id].lastUpdate != 0, ErrorsLib.MARKET_NOT_CREATED);
         require(UtilsLib.exactlyOneZero(assets, shares), ErrorsLib.INCONSISTENT_INPUT);
-        // No need to verify that onBehalf != address(0) thanks to the authorization check.
         require(receiver != address(0), ErrorsLib.ZERO_ADDRESS);
+        // No need to verify that onBehalf != address(0) thanks to the following authorization check.
         require(_isSenderAuthorized(onBehalf), ErrorsLib.UNAUTHORIZED);
```
MathisGD, line 199: to me we shouldn't change this, because the comment is in the place where you expect the check (based on the functions where the check is)

src/interfaces/IOracle.sol
```diff
@@ -5,6 +5,7 @@ pragma solidity >=0.5.0;
 /// @author Morpho Labs
 /// @custom:contact security@morpho.org
 /// @notice Interface that oracles used by Morpho must implement.
+/// @dev It is the responsibility of the oracle to handle stale prices and price manipulations.
```
MathisGD, line 8: and a lot of other things then, I don't think that it has its place here
```suggestion
```

src/libraries/MathLib.sol
MathisGD: note that this exists 
image

src/Morpho.sol
MerlinEgalite: It works only for those who know very well the code like us. Not for others. So I'm against reverting the change

src/interfaces/IOracle.sol
MerlinEgalite: I think it has since we'll have other comments like this in the contest

MathisGD: Then we can answer the same thing. Trying to cover the dependencies with more depth is just dangerous for me because we cannot have a 100% coverage and people will ask to be paid for the things that we didn't cover. Still against adding such thing.

MerlinEgalite: Yeah but I'm spending my time answering the same questions several times, so why not writing the answer somewhere?

>  is just dangerous 

I don't think it is

src/interfaces/IMorpho.sol
```diff
@@ -67,12 +67,18 @@ interface IMorpho {
     function feeRecipient() external view returns (address);
 
     /// @notice The state of the position of `user` on the market corresponding to `id`.
+    /// @dev Warning: `supplyShares` does not contain the additional shares accrued by `feeRecipient` since the last
+    /// interest accrual.
```
MathisGD, line 71:
```suggestion
    /// @dev Warning: For `feeRecipient`, `supplyShares` does not contain the additional shares accrued since the last
    /// interest accrual.
```

src/interfaces/IOracle.sol
MathisGD: No but we can find other solutions. We can for example make it clearer that this is not an issue, by acknowledging that nothing more is expected from the oracles and that it's the users responsibility to select market with oracles that are safe. Without ever precising what "safe" means.

Jean-Grimal (approved):

Rubilmax (dismissed):

src/interfaces/IIrm.sol
```diff
@@ -9,12 +9,10 @@ import {MarketParams, Market} from "./IMorpho.sol";
 /// @notice Interface that Interest Rate Models (IRMs) used by Morpho must implement.
 interface IIrm {
     /// @notice Returns the borrow rate of the market `marketParams`.
-    /// @param marketParams The MarketParams struct of the market.
-    /// @param market The Market struct of the market.
+    /// @dev Assumes that `market` is related to the correct `marketParams`.
```
Rubilmax, line 12:
```suggestion
    /// @dev Assumes that `market` corresponds to `marketParams`.
```

src/interfaces/IOracle.sol
```diff
@@ -5,6 +5,7 @@ pragma solidity >=0.5.0;
 /// @author Morpho Labs
 /// @custom:contact security@morpho.org
 /// @notice Interface that oracles used by Morpho must implement.
+/// @dev It is the user responsibility to select markets with oracles that are safe.
```
Rubilmax, line 8:
```suggestion
/// @dev It is the user's responsibility to select markets with safe oracles.
```

Jean-Grimal (approved):

src/interfaces/IIrm.sol
```diff
@@ -9,12 +9,10 @@ import {MarketParams, Market} from "./IMorpho.sol";
...
     /// @notice Returns the borrow rate of the market `marketParams`.
-    /// @param marketParams The MarketParams struct of the market.
-    /// @param market The Market struct of the market.
+    /// @dev Assumes that `market` corresponds to `marketParams`.
     function borrowRate(MarketParams memory marketParams, Market memory market) external returns (uint256);
 
     /// @notice Returns the borrow rate of the market `marketParams` without modifying any storage.
-    /// @param marketParams The MarketParams struct of the market.
-    /// @param market The Market struct of the market.
+    /// @dev Assumes that `market` is related to the correct `marketParams`.
```
Rubilmax, line 16:
```suggestion
    /// @dev Assumes that `market` correponds to `marketParams`.
```

MerlinEgalite:
there was a typo in fact in your suggestion @Rubilmax ...

MathisGD (approved):

Jean-Grimal (dismissed):

Rubilmax (dismissed):

Rubilmax (dismissed):

MathisGD (dismissed):

Rubilmax (dismissed):

Jean-Grimal (approved):

QGarchery (dismissed):

QGarchery:
Resolving the conflict will be awkward :/
Because we duplicated the interfaces, we will now have to duplicate the comments in both of the functions (on top of the comments on the structs)

src/interfaces/IMorpho.sol
```diff
@@ -319,8 +326,14 @@ interface IMorphoStaticTyping is IMorphoBase {
 /// @dev Use this interface for Morpho to have access to all the functions with the appropriate function signatures.
 interface IMorpho is IMorphoBase {
     /// @notice The state of the position of `user` on the market corresponding to `id`.
+    /// @dev Warning: For `feeRecipient`, `supplyShares` does not contain the additional shares accrued since the last
+    /// interest accrual.
```
QGarchery, line 330:
```suggestion
    /// @dev Warning: For `feeRecipient`, `supplyShares` does not contain the accrued shares since the last interest
    /// accrual.
```
to have the same comment as on the struct (this is showing why it's not good practice to duplicate comments either)

MerlinEgalite: > to have the same comment as on the struct (this is showing why it's not good practice to duplicate comments either)

Even on an immutable contract?

Rubilmax (approved):

MerlinEgalite:
I'm not sure why the test is failing. Is it related to https://github.com/morpho-org/morpho-blue/pull/576 @Rubilmax ?

MathisGD (approved):

PR #560, opened 2023-10-24
Feat/add npm ci

julien-devatom:
## Motivation
Morpho blue does not have any dependency (forge or npm). We can use the contracts in both `hardhat` and `foundry` projects. 
If we want to ease Blue integrations to devs not working on Foundry, we can publish the package to npm.

However, publication to npm is a critical action that needs to have security about when and who can trigger the CI.

## Solution
- The workflow can be triggered by hand.
- 've not defined an auto versioning not to have to allow the action to push on main the bumped version (security reason).
- However, it is up to the person triggering the workflow to define a correct version. 
- There is a check to verify the semver format.
- I have created a GitHub environment (named npm) that is scoping the deployment to the main branch only and that requires administrators' reviewals before publishing on npm

### DIFF
--- a/.github/workflows/npm-release.yml
+++ b/.github/workflows/npm-release.yml
@@ -2,11 +2,6 @@ name: Publish on NPM
 
 on:
   workflow_dispatch:
-    inputs:
-      version_type:
-        description: 'Version to release, in valid semver format'
-        required: true
-        type: string
 
 jobs:
   publish-to-npm:
@@ -16,21 +11,8 @@ jobs:
       name: npm
       url: https://www.npmjs.com/package/@morpho-org/morpho-blue
     steps:
-
-      - name: Validate semver format of the provided version
-        run: |
-          if [[ ! "${{ github.event.inputs.version_type }}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-              echo "Invalid semver format. Please provide a valid semver format (v1.2.3)"
-              exit 1
-          fi
-
       - name: Checkout
         uses: actions/checkout@v3
-
-      - name: Bump package.json version
-        run: |
-          version_type=$(echo "${{ steps.determine_version.outputs.version_type }}")
-          yarn version --new-version ${version_type} --no-git-tag-version
 
       - name: Publish to npm
         run: |
--- a/.github/workflows/npm-release.yml
+++ b/.github/workflows/npm-release.yml
@@ -8,15 +8,13 @@ on:
         required: true
         type: string
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
-  cancel-in-progress: true
-
-
 jobs:
   publish-to-npm:
     name: Publish to NPM
     runs-on: ubuntu-latest
+    environment:
+      name: npm
+      url: https://www.npmjs.com/package/@morpho-org/morpho-blue
     steps:
 
       - name: Validate semver format of the provided version
--- a/.github/workflows/npm-release.yml
+++ b/.github/workflows/npm-release.yml
@@ -1,0 +1,40 @@
+name: Publish on NPM
+
+on:
+  workflow_dispatch:
+    inputs:
+      version_type:
+        description: 'Version to release, in valid semver format'
+        required: true
+        type: string
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
+  cancel-in-progress: true
+
+
+jobs:
+  publish-to-npm:
+    name: Publish to NPM
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Validate semver format of the provided version
+        run: |
+          if [[ ! "${{ github.event.inputs.version_type }}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+              echo "Invalid semver format. Please provide a valid semver format (v1.2.3)"
+              exit 1
+          fi
+
+      - name: Checkout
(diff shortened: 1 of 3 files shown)


MathisGD (dismissed):

package.json
```diff
@@ -1,8 +1,13 @@
 {
-  "name": "morpho-blue",
+  "name": "@morpho-org/morpho-blue",
   "description": "Morpho Blue Protocol",
   "license": "BUSL-1.1",
-  "version": "1.0.0",
+  "version": "v1.0.0",
```
Rubilmax, line 5: Resources:
- https://docs.npmjs.com/cli/v10/configuring-npm/package-json#version
- https://github.com/npm/node-semver

It seems this version **might** work. Why aren't we using `1.0.0` instead?

package.json
```diff
@@ -1,8 +1,13 @@
...
+  "files": [
+    "src",
+    "README.md",
+    "LICENSE"
+  ],
```
Rubilmax, line 10: Can we clarify why `lib` is included in the package bundle before merging this pls?

package.json
Rubilmax: Is it the syntax of `yarn version`??

julien-devatom: Yes, this is the default syntax used by version. Here we are using hardcoded version, but I've reused the default behavior of --patch, --minor & --major

package.json
julien-devatom: A bit weird since its working well on this repo..

image

package.json
Rubilmax: If I do it myself, I can increment versions using `yarn version --patch/minor/major` and it will keep the syntax `x.x.x` instead of `vx.x.x`

.github/workflows/npm-release.yml
```diff
@@ -0,0 +1,38 @@
...
+              exit 1
+          fi
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Bump package.json version
+        run: |
+          version_type=$(echo "${{ steps.determine_version.outputs.version_type }}")
+          yarn version --new-version ${version_type} --no-git-tag-version
```
Rubilmax, line 33: The `--new-version` parameter seems to be available with `yarn publish`: https://classic.yarnpkg.com/lang/en/docs/cli/publish/#toc-yarn-publish-new-version

.github/workflows/npm-release.yml
```diff
@@ -0,0 +1,38 @@
+name: Publish on NPM
+
+on:
+  workflow_dispatch:
+    inputs:
+      version_type:
```
Rubilmax, line 6: It seems you expect as input a specific version and not a version type (patch,minor,major). Is it expected? Why not use incremental versioning, considering versions cannot be republished on npm?

.github/workflows/npm-release.yml
Rubilmax: Just tested it: it doesn't commit it to git
But it's actually a benefit because it doesn't break the commit msg syntax convention that we use (whereas `yarn version` commits `vx.x.x` as commit)
So it means using `yarn publish --new-version` allows to publish the new version and optionally commit it with a defined commit msg

Also, committing the version **after** having published it seems better because in case the `publish` step fails, no git tag/commit is pushed

Rubilmax (changes requested):

package.json
Rubilmax: You don't have to publish a package to npm to know this ahah: `yarn pack`

Rubilmax: Alright I confirm so this discussion can be resolved

Rubilmax:
Just in case we ever publish ts: https://github.com/morpho-org/morpho-blue/pull/561

.github/workflows/npm-release.yml
Rubilmax: I am surprised they recommend to push the commit right after versioning: https://classic.yarnpkg.com/lang/en/docs/cli/version/#toc-version-lifecycle

But I guess it only depends on what we want. So no strict requirement on this from my end in the end

I also discover: `yarn config set version-git-message` which I think is important though, for our commit msg convention

Rubilmax (changes requested):

package.json
julien-devatom: Wanted to reproduce the exact same flow (idk if publish has some extra things that pack has not), btw Ive just deleted the package 😆

.github/workflows/npm-release.yml
julien-devatom: This is written in the PR description: 've not defined an auto versioning not to have to allow the action to push on main the bumped version (security reason).

Autoincrement is based on the version provided inside of the package.json, so the CI need to push the bumped version on main, which is not a good thing for a critical repo like this one IMO.

.github/workflows/npm-release.yml
Rubilmax: Isn't there a way to tell `yarn` to not version anything and just use the package's version defined in `package.json`?

So the flow would be:
1. commit a new version to `package.json`
2. merge
3. trigger action without any input
4. if action is triggered twice with the same version, fail

julien-devatom: That's a good tradeoff. Ive also reduced the version to 0.1.0, and we'll release the 1.0.0 once deployed (it should never change after)

Rubilmax (approved):

.github/workflows/npm-release.yml
Rubilmax: This discussion can be resolved but I keep it open for others to know

MerlinEgalite (approved):

Jean-Grimal (approved):

PR #567, opened 2023-10-26
Refactor interfaces

### LINKED ISSUE #555: Expose structs in `IMorpho` by decomposing layers of interfaces
Similar to what we've done in another private repository which I can't talk about publicly 👀 👀 👀

MerlinEgalite:
- Fixes #555 
- Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/51

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -1,7 +1,16 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.19;
 
-import {Id, IMorpho, MarketParams, Position, Market, Authorization, Signature} from "./interfaces/IMorpho.sol";
+import {
+    Id,
+    IMorphoStaticTyping,
+    IMorphoBase,
+    MarketParams,
+    Position,
+    Market,
+    Authorization,
+    Signature
+} from "./interfaces/IMorpho.sol";
 import {
     IMorphoLiquidateCallback,
     IMorphoRepayCallback,
@@ -26,7 +35,7 @@ import {SafeTransferLib} from "./libraries/SafeTransferLib.sol";
 /// @author Morpho Labs
 /// @custom:contact security@morpho.org
 /// @notice The Morpho contract.
-contract Morpho is IMorpho {
+contract Morpho is IMorphoStaticTyping {
     using MathLib for uint128;
     using MathLib for uint256;
     using UtilsLib for uint256;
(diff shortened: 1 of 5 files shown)


src/interfaces/IMorpho.sol
```diff
@@ -298,3 +280,41 @@ interface IMorpho {
...
+            uint128 totalBorrowAssets,
+            uint128 totalBorrowShares,
+            uint128 lastUpdate,
+            uint128 fee
+        );
+}
+
+/// @title IMorpho
+/// @author Morpho Labs
+/// @custom:contact security@morpho.org
```
Rubilmax, line 312: We didn't add this in `metamorpho`

Should we?

Rubilmax (approved):

src/interfaces/IMorpho.sol
QGarchery: I don't know, we are kind of breaking one of our convention because we don't have one library/interface/contract per file. I think it's fine because we want those to be "internal", in the sense that they are not supposed to be used outside of this file. Because of this, now that I think about it, I would just add this header to `IMorpho`, the interface that is actually meant to be used

QGarchery (dismissed):

MathisGD (commented):
asked a question in the issue

Jean-Grimal (approved):

src/Morpho.sol
```diff
@@ -26,7 +35,7 @@ import {SafeTransferLib} from "./libraries/SafeTransferLib.sol";
 /// @author Morpho Labs
 /// @custom:contact security@morpho.org
 /// @notice The Morpho contract.
-contract Morpho is IMorpho {
+contract Morpho is IMorphoStaticTyping {
```
MathisGD, line 38: it's weird t not have `Morpho is IMorpho`

I would have named IMorpho something like IMorphoStructs and IMorphoStaticTyping IMorpho instead

src/Morpho.sol
```diff
@@ -1,7 +1,16 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity 0.8.19;
 
-import {Id, IMorpho, MarketParams, Position, Market, Authorization, Signature} from "./interfaces/IMorpho.sol";
+import {
+    Id,
+    IMorphoStaticTyping,
+    IMorphoBase,
```
MathisGD, line 7: why do we import IMorphoBase ? doesn't IMorphoStaticTyping inherit from IMorphoBase?

MerlinEgalite: yes it does but the compiler does not allow to use `IMorphoStaticTyping` only...

src/Morpho.sol
Rubilmax: I prefer having it unnatural for readers and natural for integrators rather than the other way around

MathisGD: normal clash, as I'm not an integrator and you are (in some sense) 😄

Rubilmax (approved):

MathisGD (approved):

Jean-Grimal (approved):

QGarchery (approved):

PR #568, opened 2023-10-26
Funds could be stuck if IRM reverts

MerlinEgalite:
Fixes https://github.com/cantinasec/review-morpho-blue-1/issues/41

### DIFF
--- a/src/interfaces/IMorpho.sol
+++ b/src/interfaces/IMorpho.sol
@@ -139,11 +139,12 @@ interface IMorpho {
     /// - The token balance of the sender (resp. receiver) should decrease (resp. increase) by exactly the given amount
     /// on `transfer` and `transferFrom`. In particular, tokens with fees on transfer are not supported.
     /// - The IRM should not re-enter Morpho.
-    /// @dev Here is a list of properties on the market's dependencies that could break Morpho's liveness properties:
+    /// @dev Here is a list of properties on the market's dependencies that could break Morpho's liveness properties
+    /// (funds could get stuck):
     /// - The token can revert on `transfer` and `transferFrom` for a reason other than an approval or balance issue.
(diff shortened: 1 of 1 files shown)


src/interfaces/IMorpho.sol
```diff
@@ -143,7 +143,7 @@ interface IMorpho {
     /// - The token can revert on `transfer` and `transferFrom` for a reason other than an approval or balance issue.
     /// - A very high amount of assets (~1e35) supplied or borrowed can make the computation of `toSharesUp` and
     /// `toSharesDown` overflow.
-    /// - The IRM can revert on `borrowRate`.
+    /// - The IRM can revert on `borrowRate` which means it could lead to funds being stuck.
```
Rubilmax, line 146: I believe it is the case for all of these liveness-breaking situations, so we should duplicate this comment everytime?

But I understand that "break Morpho's liveness properties" imply that funds may get stuck in some cases, so I would say this comment is simply highlighting that no liveness => funds may be stuck
Perhaps we want to change the wording of the sentence introducing the bullet points?

MerlinEgalite: That was an argument made in the related issue. We can acknowledge it also

QGarchery: Agree with @Rubilmax, I would just add that funds **could** get stuck after "break Morpho's liveness properties"

Rubilmax: Ah, right: https://github.com/cantinasec/review-morpho-blue-1/issues/41#issuecomment-1780862265
I'm fine with both!

Rubilmax (approved):

QGarchery (approved):

Jean-Grimal (dismissed):

MathisGD (approved):

QGarchery (approved):

PR #569, opened 2023-10-26
fix(morpho): remove unnecessary casts

### LINKED ISSUE #566: Unnecessary cast
There's also another unnecessary usage of `toUint128` at L372

_Originally posted by @Rubilmax in https://github.com/morpho-org/morpho-blue/pull/557#discussion_r1371725357_

Rubilmax:
- Fixes #566

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -30,6 +30,7 @@ contract Morpho is IMorpho {
     using MathLib for uint128;
     using MathLib for uint256;
     using UtilsLib for uint256;
+    using SharesMathLib for uint128;
     using SharesMathLib for uint256;
     using SafeTransferLib for IERC20;
     using MarketParamsLib for MarketParams;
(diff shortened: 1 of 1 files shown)


QGarchery (approved):
Clean

MerlinEgalite (approved):

Jean-Grimal (approved):

src/Morpho.sol
```diff
@@ -272,7 +273,8 @@ contract Morpho is IMorpho {
 
         position[id][onBehalf].borrowShares -= shares.toUint128();
         market[id].totalBorrowShares -= shares.toUint128();
-        market[id].totalBorrowAssets = UtilsLib.zeroFloorSub(market[id].totalBorrowAssets, assets).toUint128();
+        // Safe "unchecked" cast since the value is smaller than totalBorrowAssets.
+        market[id].totalBorrowAssets = uint128(UtilsLib.zeroFloorSub(market[id].totalBorrowAssets, assets));
```
MathisGD, line 277: I don't think that we should add think kind of optimization at this stage (even earlier in the development actually), similarly to #556. We don't know how much gas does it save right? (no veto though, it's correct indeed).

src/Morpho.sol
```diff
@@ -369,12 +371,13 @@ contract Morpho is IMorpho {
 
         position[id][borrower].borrowShares -= repaidShares.toUint128();
         market[id].totalBorrowShares -= repaidShares.toUint128();
-        market[id].totalBorrowAssets = UtilsLib.zeroFloorSub(market[id].totalBorrowAssets, repaidAssets).toUint128();
+        // Safe "unchecked" cast since the value is smaller than totalBorrowAssets.
+        market[id].totalBorrowAssets = uint128(UtilsLib.zeroFloorSub(market[id].totalBorrowAssets, repaidAssets));
```
MathisGD, line 375: same

MathisGD (changes requested):

src/Morpho.sol
Rubilmax: I unfortunately agree that this PR has the exact same stakes than #556 so we shouldn't merge it... Too bad we realize late than we've complexified our lives

MerlinEgalite:
Closing because we don't want to introduce small optimizations at this stage that we usually dismiss

PR #573, opened 2023-11-03
removed unchecked casts

MathisGD:
before

image

after

image

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -142,7 +142,6 @@ contract Morpho is IMorpho {
         require(isLltvEnabled[marketParams.lltv], ErrorsLib.LLTV_NOT_ENABLED);
         require(market[id].lastUpdate == 0, ErrorsLib.MARKET_ALREADY_CREATED);
 
-        // Safe "unchecked" cast.
         market[id].lastUpdate = block.timestamp.toUint128();
         idToMarketParams[id] = marketParams;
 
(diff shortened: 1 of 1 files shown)


src/Morpho.sol
```diff
@@ -479,7 +476,7 @@ contract Morpho is IMorpho {
         }
 
         // Safe "unchecked" cast.
```
Rubilmax, line 478: This comment can be removed

src/Morpho.sol
```diff
@@ -144,7 +143,7 @@ contract Morpho is IMorpho {
         require(market[id].lastUpdate == 0, ErrorsLib.MARKET_ALREADY_CREATED);
 
         // Safe "unchecked" cast.
```
Rubilmax, line 145: This comment can be removed

src/Morpho.sol
```diff
@@ -378,15 +377,13 @@ contract Morpho is IMorpho {
         if (position[id][borrower].collateral == 0) {
             badDebtShares = position[id][borrower].borrowShares;
             // Safe "unchecked" cast since the value is smaller than totalBorrowAssets.
```
Rubilmax, line 379: This comment can be removed

Rubilmax (commented):
I'm not strongly against as I understand we're moving to a "safer" code but strictly speaking it's just adding code that we know from reasoning will never be actually used (so we're adding some dead code)

So it goes against minimalism to me ^^

src/Morpho.sol
MathisGD:
```suggestion
```

src/Morpho.sol
MathisGD:
```suggestion
```

Issue #460, opened 2023-08-31
Never skip interest accrual ?


MathisGD:
Currently, interest accrual is skipped if 1) elapsed == 0 or if 2) totalBorrowAssets = 0.

https://github.com/morpho-labs/morpho-blue/blob/60e6af136403962ec18fef98b3fbf49d064c11e6/src/Morpho.sol#L450-L452

This is is sort of problematic for IRMs, because 1) it increases the complexity of building an IRM (you have to think about this) and 2) it reduces the possibilities with your IRM. What if you wanted to do something particular when the totalSupply is zero?

Also, it makes sense to let this IRM manage this "opti" instead of doing it inside of Blue. Wdyt ?

PR #574, opened 2023-11-04
Remove IRM call skip

MathisGD:
I think that we should remove the IRM call skip bacause (https://github.com/morpho-org/morpho-blue/issues/460):
- it makes building IRMs more complex (the dev has to think about this)
- it creates some weird cases in our first IRM (https://github.com/morpho-labs/morpho-blue-irm/pull/54#discussion_r1380565416)
- we have in mind not so crazy scenarios where one might want to do something particular when a market stayed empty of with only supply for a certain period of time (that actually includes my second point) (https://github.com/morpho-org/morpho-blue/issues/460#issue-1875682799)
- it simplifies Morpho Blue
- we don't care about this opti, interacting with a market without any borrow is super rare
- https://github.com/cantinasec/review-morpho-blue-1/issues/42

### DIFF
--- a/src/libraries/periphery/MorphoBalancesLib.sol
+++ b/src/libraries/periphery/MorphoBalancesLib.sol
@@ -45,7 +45,7 @@ library MorphoBalancesLib {
 
         uint256 elapsed = block.timestamp - market.lastUpdate;
 
-        if (elapsed != 0) {
+        if (elapsed != 0 && market.totalBorrowAssets != 0) {
             uint256 borrowRate = IIrm(marketParams.irm).borrowRateView(marketParams, market);
             uint256 interest = market.totalBorrowAssets.wMulDown(borrowRate.wTaylorCompounded(elapsed));
             market.totalBorrowAssets += interest.toUint128();
--- a/src/libraries/periphery/MorphoBalancesLib.sol
+++ b/src/libraries/periphery/MorphoBalancesLib.sol
@@ -45,6 +45,7 @@ library MorphoBalancesLib {
 
         uint256 elapsed = block.timestamp - market.lastUpdate;
 
+        // Skipped if elapsed == 0 of if totalBorrowAssets == 0 because interest would be null.
         if (elapsed != 0 && market.totalBorrowAssets != 0) {
             uint256 borrowRate = IIrm(marketParams.irm).borrowRateView(marketParams, market);
             uint256 interest = market.totalBorrowAssets.wMulDown(borrowRate.wTaylorCompounded(elapsed));
--- a/src/libraries/periphery/MorphoBalancesLib.sol
+++ b/src/libraries/periphery/MorphoBalancesLib.sol
@@ -45,7 +45,7 @@ library MorphoBalancesLib {
 
         uint256 elapsed = block.timestamp - market.lastUpdate;
 
-        if (elapsed != 0 && market.totalBorrowAssets != 0) {
+        if (elapsed != 0) {
             uint256 borrowRate = IIrm(marketParams.irm).borrowRateView(marketParams, market);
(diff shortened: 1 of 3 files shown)


MerlinEgalite:
I restarted the CI but it seems very very slow, I'm not sure why

MerlinEgalite:
I think we should link this PR to https://github.com/cantinasec/review-morpho-blue-1/issues/42 btw

Rubilmax (approved):

src/mocks/IrmMock.sol
```diff
@@ -10,6 +10,7 @@ contract IrmMock is IIrm {
     using MathLib for uint128;
 
     function borrowRateView(MarketParams memory, Market memory market) public pure returns (uint256) {
+        if (market.totalSupplyAssets == 0) return 0;
         uint256 utilization = market.totalBorrowAssets.wDivDown(market.totalSupplyAssets);
```
Rubilmax, line 14:
```suggestion
        if (market.totalSupplyAssets == 0) return 0;

        uint256 utilization = market.totalBorrowAssets.wDivDown(market.totalSupplyAssets);
```

non-blocking

Rubilmax (approved):

Jean-Grimal (approved):

Rubilmax (changes requested):
The check should be removed from libraries too

QGarchery:
Talking with @Rubilmax and @MerlinEgalite we found out that removing this check also makes it more complex to enable markets with no borrow, such as the idle market that we consider using in metamorpho. It could still be done by enabling a dummy IRM that does nothing (so it never reverts and doesn't cost much gas to call), but it becomes awkward to do

Rubilmax:
> note that it's not mandatory as it would give the same result (as we use borrowRateView here, we don't care about calling it or not)

That seems correct!

src/libraries/periphery/MorphoBalancesLib.sol
```diff
@@ -45,7 +45,7 @@ library MorphoBalancesLib {
 
         uint256 elapsed = block.timestamp - market.lastUpdate;
 
-        if (elapsed != 0 && market.totalBorrowAssets != 0) {
+        if (elapsed != 0) {
```
MathisGD, line 48: note that this change was not necessary, the code is correct without. But @Rubilmax finds it better like that, and I'm ok.

Jean-Grimal (approved):

MerlinEgalite (dismissed):
It must be linked to the relevant cantina issue

MathisGD:
> It must be linked to the relevant cantina issue

done

src/libraries/periphery/MorphoBalancesLib.sol
MathisGD: >This change is required if the rate is non-zero while totalBorrowAssets is zero. Could such an IRM exist? I don't think it makes any sense, so we may prefer to skip the view call in this case and save some gas?

no it's not true, because in Blue

```solidity
uint256 interest = market.totalBorrowAssets.wMulDown(borrowRate.wTaylorCompounded(elapsed));
```

so the interest is always zero when there is no borrow

Rubilmax: Alright then I agree to revert this change ^^

Rubilmax (dismissed):

Jean-Grimal (dismissed):

QGarchery (approved):

src/libraries/periphery/MorphoBalancesLib.sol
MerlinEgalite: I think we should add a comment because it's not obvious at all why we're doing this here

MerlinEgalite (approved):

Jean-Grimal (approved):

QGarchery (approved):

PR #576, opened 2023-11-06
ci(foundry): skip failing invariants

Rubilmax:


MathisGD:
I don't think that this is necessary, but we have tests really failing on invariants

Rubilmax:
> I don't think that this is necessary, but we have tests really failing on invariants

Pls run the invariant test suite locally to see why it's necessary ^^

MathisGD (dismissed):

MerlinEgalite (approved):

PR #592, opened 2023-11-12
docs: document inflation attack (spearbit 40)

MathisGD:
https://github.com/cantinasec/review-morpho-blue-1/issues/40

### DIFF
--- a/src/interfaces/IMorpho.sol
+++ b/src/interfaces/IMorpho.sol
@@ -124,8 +124,7 @@ interface IMorphoBase {
     /// - A very high price returned by the oracle can make the computation of `maxBorrow` in `_isHealthy` overflow, or
     /// the computation of `assetsRepaid` in `liquidate` overflow.
     /// @dev The borrow share price of a market with less than 1e4 assets borrowed can be decreased by manipulations, to
-    /// the
-    /// point where `totalBorrowShares` is very large and borrowing overflows.
+    /// the point where `totalBorrowShares` is very large and borrowing overflows.
     function createMarket(MarketParams memory marketParams) external;
 
     /// @notice Supplies `assets` or `shares` on behalf of `onBehalf`, optionally calling back the caller's
--- a/src/interfaces/IMorpho.sol
+++ b/src/interfaces/IMorpho.sol
@@ -123,7 +123,8 @@ interface IMorphoBase {
     /// `liquidate` from being used under certain market conditions.
     /// - A very high price returned by the oracle can make the computation of `maxBorrow` in `_isHealthy` overflow, or
     /// the computation of `assetsRepaid` in `liquidate` overflow.
-    /// @dev The borrow share price of a market with less than 1e4 assets borrowed can be decreased by manipulations, to the
+    /// @dev The borrow share price of a market with less than 1e4 assets borrowed can be decreased by manipulations, to
+    /// the
     /// point where `totalBorrowShares` is very large and borrowing overflows.
     function createMarket(MarketParams memory marketParams) external;
 
--- a/src/interfaces/IMorpho.sol
+++ b/src/interfaces/IMorpho.sol
@@ -123,8 +123,8 @@ interface IMorphoBase {
     /// `liquidate` from being used under certain market conditions.
     /// - A very high price returned by the oracle can make the computation of `maxBorrow` in `_isHealthy` overflow, or
     /// the computation of `assetsRepaid` in `liquidate` overflow.
-    /// @dev The share price of a market with less than 1e4 assets borrowed can be decreased by manipulations, to the
-    /// point where totalBorrowShares is very large and borrowing overflows.
+    /// @dev The borrow share price of a market with less than 1e4 assets borrowed can be decreased by manipulations, to the
+    /// point where `totalBorrowShares` is very large and borrowing overflows.
     function createMarket(MarketParams memory marketParams) external;
 
     /// @notice Supplies `assets` or `shares` on behalf of `onBehalf`, optionally calling back the caller's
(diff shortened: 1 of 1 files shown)


src/interfaces/IMorpho.sol
```diff
@@ -123,6 +123,8 @@ interface IMorphoBase {
     /// `liquidate` from being used under certain market conditions.
     /// - A very high price returned by the oracle can make the computation of `maxBorrow` in `_isHealthy` overflow, or
     /// the computation of `assetsRepaid` in `liquidate` overflow.
+    /// @dev For a market with less than 1e4 assets borrowed in total, a manipulation is possible to inflate the borrow
+    /// shares' price to the point where borrower can no longer interact with the market.
```
Rubilmax, line 127:
```suggestion
    /// @dev The share price of a market with less than 1e4 total borrowed assets can be deflated to a point that borrowing is no longer possible.
```

I am fine with replacing "deflated" with "decreased"

In any case, I don't think that "share price inflation" is the right choice of words: the goal of the attack is to inflate total borrow shares, so deflate the share price

Right?

Rubilmax (changes requested):

src/interfaces/IMorpho.sol
Rubilmax: I am also fine with replacing "borrowing is no longer possible" with "borrowing overflows" or "borrowing reverts"

Rubilmax: Or, to a broader extent, that "can be freely deflated / decreased."

To indicate that the issue is not only about DoSing a market, but also manipulating the share price

MerlinEgalite (approved):

src/interfaces/IMorpho.sol
```diff
@@ -123,6 +123,8 @@ interface IMorphoBase {
     /// `liquidate` from being used under certain market conditions.
     /// - A very high price returned by the oracle can make the computation of `maxBorrow` in `_isHealthy` overflow, or
     /// the computation of `assetsRepaid` in `liquidate` overflow.
+    /// @dev The share price of a market with less than 1e4 assets borrowed can be increased by manipulations, to the
```
Rubilmax, line 126: I still don't get it, isn't the attack about inflating the total borrow shares, so deflating / decreasing the share price?

Rubilmax (changes requested):

src/interfaces/IMorpho.sol
```diff
@@ -123,6 +123,8 @@ interface IMorphoBase {
...
+    /// point where borrowing and repaying overflows.
```
Rubilmax, line 127: Repaying can overflow?

MathisGD: yes because it uses totalBorrowShares almost the same way as borrow

src/interfaces/IMorpho.sol
MathisGD: you are right, I'm dumb

src/interfaces/IMorpho.sol
Rubilmax: But how can total borrow shares overflow if it's only decreased upon repay?
I think I'm missing something

MathisGD: my bad, I thought that the overflow happened during a WAD mul

MerlinEgalite (dismissed):

Rubilmax (approved):

src/interfaces/IMorpho.sol
```diff
@@ -123,6 +123,8 @@ interface IMorphoBase {
     /// `liquidate` from being used under certain market conditions.
     /// - A very high price returned by the oracle can make the computation of `maxBorrow` in `_isHealthy` overflow, or
     /// the computation of `assetsRepaid` in `liquidate` overflow.
+    /// @dev The share price of a market with less than 1e4 assets borrowed can be decreased by manipulations, to the
+    /// point where totalBorrowShares is very large and borrowing overflows.
```
QGarchery, line 127:
```suggestion
    /// @dev When there are less than 1e4 assets borrowed in a market, their share price can be decreased by
    /// manipulations, to the point where `totalBorrowShares` is very large and borrowing overflows.
```
because share price can also refer to the share price of supply assets

Jean-Grimal (commented):
Maybe we should add a comment explaining that this can be avoiding within the first borrow on the market

src/interfaces/IMorpho.sol
Rubilmax:
```suggestion
    /// @dev The borrow share price of a market with less than 1e4 assets borrowed can be decreased by manipulations, to the
    /// point where totalBorrowShares is very large and borrowing overflows.
```

MathisGD:
> Maybe we should add a comment explaining that this can be avoiding within the first borrow on the market

To me it's enough like that. Let's keep comments minimal.

src/interfaces/IMorpho.sol
```diff
@@ -123,6 +123,9 @@ interface IMorphoBase {
     /// `liquidate` from being used under certain market conditions.
     /// - A very high price returned by the oracle can make the computation of `maxBorrow` in `_isHealthy` overflow, or
     /// the computation of `assetsRepaid` in `liquidate` overflow.
+    /// @dev The borrow share price of a market with less than 1e4 assets borrowed can be decreased by manipulations, to
+    /// the
+    /// point where `totalBorrowShares` is very large and borrowing overflows.
```
QGarchery, line 128:
```suggestion
    /// the point where `totalBorrowShares` is very large and borrowing overflows.
```

Jean-Grimal:
> > Maybe we should add a comment explaining that this can be avoiding within the first borrow on the market
> 
> To me it's enough like that. Let's keep comments minimal.

Actually you are right, and my point is not true, because as mentioned in #593 the share price might be reset anyway if at some point there is no borrow

QGarchery:
> > > Maybe we should add a comment explaining that this can be avoiding within the first borrow on the market
> > 
> > 
> > To me it's enough like that. Let's keep comments minimal.
> 
> Actually you are right, and my point is not true, because as mentioned in #593 the share price might be reset anyway if at some point there is no borrow

Yes and there is no practical way to ensure that the borrow stays on the contract. The usual method of doing a small transaction and forgetting the key does not work, because that small transaction is a borrow and you might get liquidated eventually

MerlinEgalite (approved):

QGarchery (approved):

PR #593, opened 2023-11-13
docs: document reset share price jump

### LINKED ISSUE #562: Slippage issue when resetting a market
Thought about this issue with @Rubilmax, in particular the idea for a general fix is his.

### Shares value in Morpho Blue

When a market becomes empty, the value of shares is resetted to the initial default rate of `VIRTUAL_ASSETS = 1` and `VIRTUAL_SHARES = 1e6`. This means that in effect an attacker can frontrun a transaction to change the value of shares. In Morpho Blue, shares value change like this:
1. value of both supply shares and borrow shares increase with accrued interest
2. value of supply shares goes up with roundings and value of borrow shares go down with roundings at each interaction. This effect is way smaller than the effect of 1
3. bad debt decreases the value of supply shares. This effect is potentially significant, and could be more than the effect of 1

Those effects have been verified with Certora [here](https://github.com/morpho-org/morpho-blue/blob/certora/dev/certora/specs/RatioMath.spec).

### Slippage issue for Morpho Blue

In practice those effects are not problematic for slippage because:
1. interests are accrued before each interaction anyway
2. this effect is negligible, we are talking about one WEI of difference
3. this effect is rare, and users are expected to lose when this happens. New suppliers should wait until bad debt is realized to ensure they incur no loss.

Resetting shares value does not have the same mitigation. Because of the above effects and their size, we can expect the value of shares to be greater than the initial value (we can discuss about this point below). So resetting decreases the shares value, which has different effects for supply and borrow:
- for the borrow side, users can take a smaller debt position than what they anticipated when they pass shares, but it does not come with the risk of entering a position too risky. Users can receive less assets than anticipated, which can make the rest of the transaction revert
- for the supply side, users can take a smaller supply position than what they anticipated when they pass shares, but it does not come with the risk of entering a position too risky because supply is not used as collateral. Users can leave more assets than anticipated on their funding account, which can potentially be lost if managed incorrectly (think of funds on the bundler contract that are then skimmed by someone else).

### General issue and potential solution

The more general issue comes with the solution that Morpho Blue has adopted to prevent shares manipulation. This solution is similar to the one of OZ in the vaults with the virtual assets and the virtual shares. This solution enforces a default ratio. Remark that enforcing a default ratio does not prevent sudden changes of ratio, which is at the core of this issue. Notice that, in the case of a vault, depositing a significant amount at creation prevents sudden changes of ratio. This is what motivates the following potential solution, by trying to reproduce a users taking significant positions right at the start:

In `createMarket`:
1. mint `VIRTUAL_SHARES` to `address(0)`
2. set `totalSupplyAssets = VIRTUAL_ASSETS` and `totalBorrowAssets = VIRTUAL_ASSETS`

Beware that the previous solution might be buggy and have unforeseen consequences

MathisGD:
Fixes #562

### DIFF
--- a/src/interfaces/IMorpho.sol
+++ b/src/interfaces/IMorpho.sol
@@ -130,7 +130,7 @@ interface IMorphoBase {
     /// @dev Either `assets` or `shares` should be zero. Most usecases should rely on `assets` as an input so the caller
     /// is guaranteed to have `assets` tokens pulled from their balance, but the possibility to mint a specific amount
     /// of shares is given for full compatibility and precision.
-    /// @dev When withdrawing all supply shares, the supply share price instantly resets to
+    /// @dev If the supply of a market gets depleted, the supply share price instantly resets to
     /// `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
     /// @dev Supplying a large amount can revert for overflow.
     /// @param marketParams The market to supply assets to.
@@ -173,7 +173,7 @@ interface IMorphoBase {
     /// @dev Either `assets` or `shares` should be zero. Most usecases should rely on `assets` as an input so the caller
     /// is guaranteed to borrow `assets` of tokens, but the possibility to mint a specific amount of shares is given for
     /// full compatibility and precision.
-    /// @dev When repaying all borrow shares, the borrow share price instantly resets to
+    /// @dev If the borrow of a market gets depleted, the borrow share price instantly resets to
     /// `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
     /// @dev `msg.sender` must be authorized to manage `onBehalf`'s positions.
     /// @dev Borrowing a large amount can revert for overflow.
--- a/src/interfaces/IMorpho.sol
+++ b/src/interfaces/IMorpho.sol
@@ -130,7 +130,7 @@ interface IMorphoBase {
     /// @dev Either `assets` or `shares` should be zero. Most usecases should rely on `assets` as an input so the caller
     /// is guaranteed to have `assets` tokens pulled from their balance, but the possibility to mint a specific amount
     /// of shares is given for full compatibility and precision.
-    /// @dev Upon reset of a market (withdrawing all supply shares), the supply share price suddenly decreases.
+/// @dev When withdrawing all supply shares, the supply share price instantly resets to `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
     /// @dev Supplying a large amount can revert for overflow.
     /// @param marketParams The market to supply assets to.
     /// @param assets The amount of assets to supply.
@@ -172,7 +172,7 @@ interface IMorphoBase {
     /// @dev Either `assets` or `shares` should be zero. Most usecases should rely on `assets` as an input so the caller
     /// is guaranteed to borrow `assets` of tokens, but the possibility to mint a specific amount of shares is given for
     /// full compatibility and precision.
-    /// @dev Upon reset of a market (repaying all borrow shares), the borrow share price suddenly decreases.
+    /// @dev When repaying all borrow shares, the borrow share price instantly resets to `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
     /// @dev `msg.sender` must be authorized to manage `onBehalf`'s positions.
     /// @dev Borrowing a large amount can revert for overflow.
     /// @param marketParams The market to borrow assets from.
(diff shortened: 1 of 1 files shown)


src/interfaces/IMorpho.sol
```diff
@@ -171,6 +172,7 @@ interface IMorphoBase {
     /// @dev Either `assets` or `shares` should be zero. Most usecases should rely on `assets` as an input so the caller
     /// is guaranteed to borrow `assets` of tokens, but the possibility to mint a specific amount of shares is given for
     /// full compatibility and precision.
+    /// @dev Upon reset of a market (repaying all borrow shares), the borrow share price suddenly decreases.
```
Rubilmax, line 175:
```suggestion
    /// @dev When repaying all borrow shares, the borrow share price instantly resets to `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
```

src/interfaces/IMorpho.sol
```diff
@@ -130,6 +130,7 @@ interface IMorphoBase {
     /// @dev Either `assets` or `shares` should be zero. Most usecases should rely on `assets` as an input so the caller
     /// is guaranteed to have `assets` tokens pulled from their balance, but the possibility to mint a specific amount
     /// of shares is given for full compatibility and precision.
+    /// @dev Upon reset of a market (withdrawing all supply shares), the supply share price suddenly decreases.
```
MathisGD, line 133:
```suggestion
/// @dev When withdrawing all supply shares, the supply share price instantly resets to `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
```

src/interfaces/IMorpho.sol
```diff
@@ -130,6 +130,7 @@ interface IMorphoBase {
     /// @dev Either `assets` or `shares` should be zero. Most usecases should rely on `assets` as an input so the caller
     /// is guaranteed to have `assets` tokens pulled from their balance, but the possibility to mint a specific amount
     /// of shares is given for full compatibility and precision.
+/// @dev When withdrawing all supply shares, the supply share price instantly resets to `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
```
Rubilmax, line 133:
```suggestion
    /// @dev When withdrawing all supply shares, the supply share price instantly resets to `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
```

Rubilmax (changes requested):

Rubilmax (approved):

Rubilmax (dismissed):

Jean-Grimal (approved):

src/interfaces/IMorpho.sol
```diff
@@ -130,6 +130,8 @@ interface IMorphoBase {
     /// @dev Either `assets` or `shares` should be zero. Most usecases should rely on `assets` as an input so the caller
     /// is guaranteed to have `assets` tokens pulled from their balance, but the possibility to mint a specific amount
     /// of shares is given for full compatibility and precision.
+    /// @dev When withdrawing all supply shares, the supply share price instantly resets to
+    /// `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
```
QGarchery, line 134:
```suggestion
    /// @dev If the supply of a market gets depleted, the supply share price instantly resets to
    /// `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
```
because:
- it's weird to only have this comment on supply and talk about withdrawing
- it can happen that there are very few shares and no assets, and in effect the ratio is reset to the initial ratio

src/interfaces/IMorpho.sol
```diff
@@ -171,6 +173,8 @@ interface IMorphoBase {
     /// @dev Either `assets` or `shares` should be zero. Most usecases should rely on `assets` as an input so the caller
     /// is guaranteed to borrow `assets` of tokens, but the possibility to mint a specific amount of shares is given for
     /// full compatibility and precision.
+    /// @dev When repaying all borrow shares, the borrow share price instantly resets to
+    /// `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
```
QGarchery, line 177:
```suggestion
    /// @dev If the borrow of a market gets depleted, the borrow share price instantly resets to
    /// `VIRTUAL_ASSETS`:`VIRTUAL_SHARES`.
```

QGarchery (approved):

MerlinEgalite (approved):

PR #601, opened 2023-11-22
feat: pass selector in transfers

MathisGD:


Rubilmax (changes requested):
I believe you need to also pass `msg.sender` as additional calldata so the permissioned token can check whether the sender is authorized (because in the ERC20's transfer context, `msg.sender` is Blue)

Or we force the ERC20 impl to rely on `tx.origin`

MathisGD:
I don't think so. The wrapper just needs to check that in the cast of a transfer from Blue to an address in the context of a liquidation, this address needs to be whitelisted. As the bundler wouldn't be whitelisted, it works

MathisGD:
I think that it's the `to` that we care about (for the collateral, and the from for the borrowable in a liquidation)

PR #604, opened 2023-11-22
[Certora] mutant 15

QGarchery:
Screenshot from 2023-11-22 19-59-22


QGarchery:
Correctly caught by AccrueInterest

PR #605, opened 2023-11-22
[Certora] mutant 14

QGarchery:
Screenshot from 2023-11-22 19-58-52


QGarchery:
Correctly caught by AccrueInterest

PR #606, opened 2023-11-22
[Certora] mutant 13

QGarchery:
Screenshot from 2023-11-22 19-58-21


QGarchery:
Warning: changed the mutant to get significant results (it was trivially failing most rules). Now it skips the require instead of reverting

PR #608, opened 2023-11-22
[Certora] mutant 11

QGarchery:
### Bad debt realization
Screenshot from 2023-11-22 19-57-10


QGarchery:
Now the mutant is also correctly caught by Liveness, thanks to #608

PR #610, opened 2023-11-22
[Certora] mutant 9

QGarchery:
Screenshot from 2023-11-22 19-55-39


QGarchery:
The reason for this formula is quite abstract, and it wouldn't make much sense to try to catch this (unit tests are better suited for it)

PR #611, opened 2023-11-22
[Certora] mutant 8

QGarchery:
Screenshot from 2023-11-22 19-55-10


QGarchery:
Caught by many rules and invariants, because it reverts all the time

QGarchery:
Warning: changed the mutant to get significant results (it was trivially failing most rules). Now it skips the require instead of reverting

QGarchery:
Correctly caught by Reverts

PR #612, opened 2023-11-22
[Certora] mutant 7

QGarchery:
Screenshot from 2023-11-22 19-54-40


QGarchery:
Correctly caught by Liveness

PR #616, opened 2023-11-22
[Certora] mutant 3

QGarchery:
Screenshot from 2023-11-22 19-52-29


QGarchery:
We could use a `satisfy` statement to make sure that the following liveness property holds: there are inputs such that you can borrow non-zero assets

PR #617, opened 2023-11-22
[Certora] mutant 2

QGarchery:
Screenshot from 2023-11-22 19-51-55


QGarchery:
Was only caught by Reverts, Liveness was improved to also catch it

Issue #390, opened 2023-08-20
Formal verification wish list


QGarchery:
List all the additional steps we could follow to improve the formal verification work:
- [x] **update to main**, notably taking into account the storage refactor (done in #373)
- [x] **fix timeouts** (mostly done in #442, see below for more)
- [x] refactor, file splitting and renaming (done in #461)
- [x] document the rules (done in #461)
- [x] document the whole verification (done in #465 and #538)
- [x] additional properties
    - [x] there are no accounts with non zero borrow and with zero collateral (done in #373, rule always collateralized)
    - [x] assuming no reverts due to external contracts, it's always possible to completely exit a borrow position, but it requires liquidity for a supply position (done for the most part in #451, except for the repayAll rule in #492)
    - [x] liveness properties about all the main functions (done in #467)
    - [x] roundtrip property with borrow-repay (done in #619)
    - [x] similar properties to the "owned/owed" properties of ExitLiquidity but for supply/borrow/supplyCollateral (see https://github.com/morpho-org/morpho-blue/pull/136#discussion_r1409030208, done in #623)
    - [x] a position gets healthier after a liquidation if its LTV was lower than 1 / LIF. This would ensure that the liquidation system has some breathing room to liquidate users before it leads to bad debt (done in #708)
    - [x] ~~check overflow reverting case on accrue interest~~ (abandoned)
    - [x] ~~check overflow reverting case on token~~ (abandoned)
    - [x] ~~check overflow reverting case on oracle~~ (abandoned)
- [x] improve rules
    - [x] `onlyLiquidateCanDecreaseSupplyRatio` can be more specific and require that the bad debt has been realized for the ratio to have decreased (new rule `liquidateWithoutBadDebtRealizationIncreasesSupplyRatio` written in #622)
    - [x] `userWithoutBorrowCannotLoseCollateral` can also state that only the liquidate function can make him lose collateral (new rule `userCannotLoseCollateralExceptLiquidate` written in https://github.com/morpho-org/morpho-blue/pull/461)
    - [x] supply-withdraw roundtrip with arbitrary withdraw (see https://github.com/morpho-org/morpho-blue/pull/345#discussion_r1297125249, done in #619)
    - [x] fix timeout on `liquidate` for the increase borrow ratio rule (done in #679)
    - [x] fix timeout on `repay` for the stay healthy rule (done in #679)
    - [x] fix timeout on `borrow` for the stay healthy rule (done in #679)
[...]

PR #623, opened 2023-12-01
[Certora] Assets accounting when entering a position

QGarchery:
This tackles one of the "additional properties" point in #390 where it is shown that users do not get more assets than what they are owed when entering a position.


certora/specs/AssetsAccounting.spec
```diff
@@ -0,0 +1,126 @@
...
+
+function ownedSupplyAssets(MorphoHarness.Id id, address user) returns uint256 {
+    uint256 userShares = supplyShares(id, user);
+    uint256 totalSupplyAssets = virtualTotalSupplyAssets(id);
+    uint256 totalSupplyShares = virtualTotalSupplyShares(id);
+
+    return libMulDivDown(userShares, totalSupplyAssets, totalSupplyShares);
+}
+
+function owedBorrowAssets(MorphoHarness.Id id, address user) returns uint256 {
```
MathisGD, line 27: this naming is weird for borrow (can make you think that these are the assets that have been transferred to you)

maybe supplyBalance and borrowBalance work good in the end

certora/specs/AssetsAccounting.spec
```diff
@@ -0,0 +1,126 @@
...
+
+    uint256 repaidAssets;
+    repaidAssets, _ = repay(e, marketParams, assets, shares, onBehalf, data);
+
+    // Assume a full repay.
+    require borrowShares(id, onBehalf) == 0;
+
+    assert repaidAssets >= owedAssets;
+}
+
```
MathisGD, line 102: we might be able to do the same rule for liquidatoin

certora/specs/AssetsAccounting.spec
```diff
@@ -0,0 +1,126 @@
...
+    MorphoHarness.Id id = libId(marketParams);
+
+    // Assume no collateral to begin with.
+    require collateral(id, onBehalf) == 0;
+
+    supplyCollateral(e, marketParams, suppliedAssets, onBehalf, data);
+
+    uint256 ownedAssets = collateral(id, onBehalf);
+
+    assert suppliedAssets >= ownedAssets;
```
MathisGD, line 114: do we want to keep the general rule?
```suggestion
    assert suppliedAssets == ownedAssets;
```

certora/specs/AssetsAccounting.spec
```diff
@@ -0,0 +1,126 @@
...
+
+// Check that the collateral assets withdrawn are less than the assets owned initially.
+rule withdrawCollateralAssetsAccounting(env e, MorphoHarness.MarketParams marketParams, uint256 withdrawnAssets, address onBehalf, address receiver) {
+    MorphoHarness.Id id = libId(marketParams);
+
+    uint256 ownedAssets = collateral(id, onBehalf);
+
+    withdrawCollateral(e, marketParams, withdrawnAssets, onBehalf, receiver);
+
+    assert withdrawnAssets <= ownedAssets;
```
MathisGD, line 125:
```suggestion
    assert withdrawnAssets == ownedAssets;
```

certora/specs/AssetsAccounting.spec
QGarchery: Ok to have something even more precise :+1:

certora/specs/AssetsAccounting.spec
QGarchery: This one won't work though, because you can withdraw less than the full amount

certora/specs/AssetsAccounting.spec
QGarchery: Not sure this is the same thing: in this file we are comparing different ways to value a position. For example, for the supply side, we compare: 
- how much it did cost you in terms of loan assets to enter a supply position
- how much you can retrieve in terms of loan assets when exiting a supply position
- an arbitrarily defined function to estimate supply assets

For the liquidations though, I don't think it defines what your borrow/collateral position is. This is just something that can happen to your position, and we could make it worse or better without impacting what you own. For example, if we had a close factor of 50%, it would not mean that your collateral is worth twice as less

certora/specs/AssetsAccounting.spec
QGarchery: I renamed it using `expectedSupplyAssets` and `expectedBorrowAssets`. It's actually computing the same thing in our case, and I feel like it's giving a good description. Just like in tests when we write `expectedValue` and compare it to the actual value

MathisGD:
let's go to main to not pollute `post-cantina` no?

PR #626, opened 2023-12-05
fix(liquidation): fix repaid assets rounding

Rubilmax:
- Fixes https://cantina.xyz/ai/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/54bbc294-2d22-4979-8cda-4db8415474f6

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -371,9 +371,13 @@ contract Morpho is IMorphoStaticTyping {
                     seizedAssets.mulDivUp(collateralPrice, ORACLE_PRICE_SCALE).wDivUp(liquidationIncentiveFactor);
                 repaidShares = repaidAssets.toSharesDown(market[id].totalBorrowAssets, market[id].totalBorrowShares);
             } else {
+                uint256 repaidAssetsDown =
+                    repaidShares.toAssetsDown(market[id].totalBorrowAssets, market[id].totalBorrowShares);
+
+                seizedAssets = repaidAssetsDown.wMulDown(liquidationIncentiveFactor).mulDivDown(
+                    ORACLE_PRICE_SCALE, collateralPrice
+                );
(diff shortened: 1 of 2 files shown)


Jean-Grimal (dismissed):

test/forge/integration/LiquidateIntegrationTest.sol
```diff
@@ -74,43 +100,6 @@ contract LiquidateIntegrationTest is BaseTest {
         uint256 lltv;
     }
 
-    function testLiquidateMargin(LiquidateTestParams memory params, uint256 amountSeized, uint256 elapsed) public {
```
QGarchery, line 77: Why removing this test ? It should pass independently on the changes made on `liquidate`

src/Morpho.sol
QGarchery, line 1: Given this new issue, I'm putting into question the importance to have the shares as input to the liquidate function. Removing this input is another (albeit radical) solution to this issue

QGarchery (changes requested):

src/Morpho.sol
Rubilmax: Besides, it seems weird to remove the granularity of repaid shares to liquidate, because a liquidation is like a repay + withdraw collateral

QGarchery: > Besides, it seems weird to remove the granularity of repaid shares to liquidate, because a liquidation is like a repay + withdraw collateral

By the same logic it should also expose a liquidate `repaidAssets` entrypoint. Ideally we would have a use case in mind to enable another entrypoint

> The only issue I have is that seizedAssets input converted into repaidShares can be greater than borrowShares thus overflow

I think it's even more general, and there are situations where the collateral seized would be less than the full collateral. Typically this is in LTV range [LLTV, 1/LIF]. The choice to have the repaid shares entrypoint was to facilitate full liquidations in those cases, and the choice was made against making bad debt realization more of a natural choice. I'm still not fully convinced by the argument in favor, and I also feel like now the code is less clear because of it

Jean-Grimal (approved):

PR #629, opened 2023-12-05
fix(safe-transfer-lib): check for code

Rubilmax:


MathisGD (approved):

Jean-Grimal (approved):

test/forge/BaseTest.sol
```diff
@@ -144,6 +144,11 @@ contract BaseTest is Test {
         return address(uint160(bound(uint256(uint160(input)), 1, type(uint160).max)));
     }
 
+    function _setValidTokens(MarketParams memory params) internal view {
+        params.loanToken = address(loanToken);
+        params.collateralToken = address(collateralToken);
+    }
+
```
QGarchery, line 151: To remove ?

QGarchery (approved):

MerlinEgalite (approved):

MathisGD (approved):

adhusson (approved):

Jean-Grimal (approved):

StErMi:
@MerlinEgalite what's the use case to allow the creation of a market that has a non-existing (yet) `loanToken` or `collateralToken` (`address(token).code.length == 0`)

src/libraries/ErrorsLib.sol
```diff
@@ -27,6 +27,9 @@ library ErrorsLib {
     /// @notice Thrown when the market is already created.
     string internal constant MARKET_ALREADY_CREATED = "market already created";
 
+    /// @notice Thrown when a token to transfer doesn't have code.
+    string internal constant NO_CODE = "no code";
```
StErMi, line 31: Nitpick: the error lib is a library shared between different contracts, consider using a more specific error name (and error message). In this case, it could be `CONTRACT_NOT_EXISTING` or `CONTRACT_NO_CODE`

MerlinEgalite:
> @MerlinEgalite what's the use case to allow the creation of a market that has a non-existing (yet) `loanToken` or `collateralToken` (`address(token).code.length == 0`)

for the collateral token there's one: create idle markets with all parameters set to 0 except the loan token

src/libraries/ErrorsLib.sol
MerlinEgalite: nitpicking goes brrrr

PR #634, opened 2023-12-06
Add call to irm at market creation

Jean-Grimal:
Fixes https://cantina.xyz/ai/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/3a235a1b-ddb4-4f7f-b254-0c940a832814

This issue doesn't really have a meaningful impact, so we could just acknowledge it as well.


QGarchery:
Updated so that it does not contain the commits on main. The CI is expected to fail on Certora because now there are other ways the `createMarket` function can revert

Rubilmax (approved):

src/Morpho.sol
```diff
@@ -158,6 +158,9 @@ contract Morpho is IMorphoStaticTyping {
         idToMarketParams[id] = marketParams;
 
         emit EventsLib.CreateMarket(id, marketParams);
+
+        // Call IRM in case it is stateful.
```
Rubilmax, line 162:
```suggestion
        // Call to initialize IRM in case it is stateful.
```

Rubilmax (approved):

src/Morpho.sol
```diff
@@ -158,6 +158,9 @@ contract Morpho is IMorphoStaticTyping {
         idToMarketParams[id] = marketParams;
 
         emit EventsLib.CreateMarket(id, marketParams);
+
+        // Call to initialize IRM in case it is stateful.
```
MerlinEgalite, line 162:
```suggestion
        // Call to initialize the IRM in case it is stateful.
```

src/Morpho.sol
```diff
@@ -158,6 +158,9 @@ contract Morpho is IMorphoStaticTyping {
         idToMarketParams[id] = marketParams;
 
         emit EventsLib.CreateMarket(id, marketParams);
+
+        // Call to initialize the IRM in case it is stateful.
+        IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
```
MathisGD, line 163: maybe it's better now to call _accrueInterest?

Rubilmax: It will be skipped because `elapsed == 0`

MathisGD: No, elapsed is very big instead (we would not have touched lastUpdated at all for the moment). In any case I agree that it's weird, let me think about a nice fix.

MathisGD: pushed in #642

MerlinEgalite:
Why do we bundle different fixes in this PR @MathisGD ?

MathisGD:
Oops it was not meant, feel free to force push

Rubilmax:
> Oops it was not meant, feel free to force push

done

MathisGD:
but my commits are still here

Rubilmax (approved):

MathisGD (approved):

PR #639, opened 2023-12-07
Document virtual borrow assets behavior

Rubilmax:


Jean-Grimal (approved):

src/libraries/SharesMathLib.sol
```diff
@@ -14,6 +14,8 @@ library SharesMathLib {
 
     /// @dev The number of virtual shares has been chosen low enough to prevent overflows, and high enough to ensure
     /// high precision computations.
+    /// @dev Warning: the assets virtual borrow shares are entitled to behave like unrealizable bad debt, but it is
```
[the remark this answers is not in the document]
Rubilmax: I don't think "expected" suits better here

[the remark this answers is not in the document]
Rubilmax: I don't get what's wrong with "entitled to": https://www.britannica.com/dictionary/eb/qa/The-Meaning-of-Entitled-To-#:~:text=Below%20are%20some%20examples%20of,the%20park%20for%20one%20month.

It's like "attitré à" un french
I don't get what the definition you share suggests either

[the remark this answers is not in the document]
MerlinEgalite: I'm still not convinced.

> To be entitled to something means you have been given the right to have or do something.

We're talking about a "right" in the case of entitle while in our case "assets virtual borrow shares" are not benefitting from a right that those shares could enjoy nor the ability to do something. They just behave like unrealizable bad debt.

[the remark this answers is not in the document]
Rubilmax: Yes, the virtual shares are actually entitled to the virtual assets at the beginning and as the value of shares grow, they are entitled to the virtual assets + some assets accrued

src/libraries/SharesMathLib.sol
```diff
@@ -14,6 +14,8 @@ library SharesMathLib {
 
     /// @dev The number of virtual shares has been chosen low enough to prevent overflows, and high enough to ensure
     /// high precision computations.
+    /// @dev Warning: The assets to which virtual borrow shares are entitled behave like unrealizable bad debt, but it
+    /// is assumed the borrow share price stays low enough to not inflate these assets.
```
MathisGD, line 18:
```suggestion
    /// is assumed the borrow share price stays low enough to not inflate these assets to something significant.
```

MerlinEgalite, line 18:
```suggestion
    /// is assumed the borrow share price stays low enough not to inflate these assets.
```

Rubilmax: suggested something very similar, tell me wyt

MathisGD (approved):

MerlinEgalite (approved):

Jean-Grimal (approved):

src/libraries/SharesMathLib.sol
```diff
@@ -14,6 +14,8 @@ library SharesMathLib {
 
     /// @dev The number of virtual shares has been chosen low enough to prevent overflows, and high enough to ensure
     /// high precision computations.
+    /// @dev Warning: The assets to which virtual borrow shares are entitled behave like unrealizable bad debt, but it
```
adhusson, line 17: entitled `to` behave?

adhusson (changes requested):

src/libraries/SharesMathLib.sol
```diff
@@ -14,8 +14,9 @@ library SharesMathLib {
 
     /// @dev The number of virtual shares has been chosen low enough to prevent overflows, and high enough to ensure
     /// high precision computations.
-    /// @dev Warning: The assets to which virtual borrow shares are entitled behave like unrealizable bad debt, but it
-    /// is assumed the borrow share price stays low enough not to inflate these assets to a significant value.
+    /// @dev Virtual shares can never be redeemed for the assets they are entitled to, but it is assumed the share price
+    /// stays low enough not to inflate these assets to a significant value.
+    /// @dev Warning: The assets to which virtual borrow shares are entitled behave like unrealizable bad debt.
```
adhusson, line 19: same here, missing `to`

MathisGD:
```suggestion
    /// @dev Warning: The assets to which virtual borrow shares are entitled to behave like unrealizable bad debt.
```

Rubilmax: The "to" is in "to which"

Rubilmax: And we argued with @MerlinEgalite this formulation was simpler that the former one: https://github.com/morpho-org/morpho-blue/pull/639#discussion_r1419245370

adhusson: ahh my bad, got it :)

adhusson (approved):

src/libraries/SharesMathLib.sol
StErMi: My 2 cents: it could be written in a more clean way? The concept you want to express/address are:

1) virtual shares (supply and borrow) cannot be redeemed (withdrawn/repaid) because they are not owned by anyone
2) virtual supply shares "steal" interest from suppliers
3) virtual borrow shares are indeed unrealizable bad debt
4) point 2 and 3 does not impact the code because "it is assumed the share price stays low enough not to inflate these assets to a significant value."

Rubilmax: I agree that's another clear way to put it
But I also think the current way is clear enough / is just fine

PR #640, opened 2023-12-07
feat(irm): skip address(0)

### LINKED ISSUE #631: IRM zero
skip IRM call (and interest accrual) when the IRM is address zero?

Rubilmax:
- Fixes #631 

it takes 85k gas to create an idle market while it takes 175k gas to create a standard market

and here is the gas diff between standard markets (left) and idle markets (right):
image

on avg, it costs -9k to supply/withdraw to/from an idle market

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -477,7 +477,7 @@ contract Morpho is IMorphoStaticTyping {
     function _accrueInterest(MarketParams memory marketParams, Id id) internal {
         uint256 elapsed = block.timestamp - market[id].lastUpdate;
 
-        if (elapsed == 0) return;
+        if (elapsed == 0 || marketParams.irm == address(0)) return;
 
         uint256 borrowRate = IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
         uint256 interest = market[id].totalBorrowAssets.wMulDown(borrowRate.wTaylorCompounded(elapsed));
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -160,7 +160,7 @@ contract Morpho is IMorphoStaticTyping {
         emit EventsLib.CreateMarket(id, marketParams);
 
         // Call to initialize the IRM in case it is stateful.
-        IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
+        if (marketParams.irm != address(0)) IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
     }
 
     /* SUPPLY MANAGEMENT */
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -475,9 +475,10 @@ contract Morpho is IMorphoStaticTyping {
     /// @dev Accrues interest for the given market `marketParams`.
     /// @dev Assumes that the inputs `marketParams` and `id` match.
     function _accrueInterest(MarketParams memory marketParams, Id id) internal {
-        uint256 elapsed = block.timestamp - market[id].lastUpdate;
+        if (marketParams.irm == address(0)) return;
 
-        if (elapsed == 0 || marketParams.irm == address(0)) return;
+        uint256 elapsed = block.timestamp - market[id].lastUpdate;
+        if (elapsed == 0) return;
 
         uint256 borrowRate = IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
         uint256 interest = market[id].totalBorrowAssets.wMulDown(borrowRate.wTaylorCompounded(elapsed));
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -483,27 +483,28 @@ contract Morpho is IMorphoStaticTyping {
     /// @dev Accrues interest for the given market `marketParams`.
     /// @dev Assumes that the inputs `marketParams` and `id` match.
     function _accrueInterest(MarketParams memory marketParams, Id id) internal {
-        if (marketParams.irm == address(0)) return;
-
         uint256 elapsed = block.timestamp - market[id].lastUpdate;
         if (elapsed == 0) return;
 
-        uint256 borrowRate = IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
-        uint256 interest = market[id].totalBorrowAssets.wMulDown(borrowRate.wTaylorCompounded(elapsed));
-        market[id].totalBorrowAssets += interest.toUint128();
-        market[id].totalSupplyAssets += interest.toUint128();
+        if (marketParams.irm != address(0)) {
+            uint256 borrowRate = IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
+            uint256 interest = market[id].totalBorrowAssets.wMulDown(borrowRate.wTaylorCompounded(elapsed));
+            market[id].totalBorrowAssets += interest.toUint128();
+            market[id].totalSupplyAssets += interest.toUint128();
 
-        uint256 feeShares;
-        if (market[id].fee != 0) {
-            uint256 feeAmount = interest.wMulDown(market[id].fee);
-            // The fee amount is subtracted from the total supply in this calculation to compensate for the fact
-            // that total supply is already increased by the full interest (including the fee amount).
-            feeShares = feeAmount.toSharesDown(market[id].totalSupplyAssets - feeAmount, market[id].totalSupplyShares);
-            position[id][feeRecipient].supplyShares += feeShares;
-            market[id].totalSupplyShares += feeShares.toUint128();
+            uint256 feeShares;
+            if (market[id].fee != 0) {
+                uint256 feeAmount = interest.wMulDown(market[id].fee);
+                // The fee amount is subtracted from the total supply in this calculation to compensate for the fact
+                // that total supply is already increased by the full interest (including the fee amount).
+                feeShares =
+                    feeAmount.toSharesDown(market[id].totalSupplyAssets - feeAmount, market[id].totalSupplyShares);
+                position[id][feeRecipient].supplyShares += feeShares;
+                market[id].totalSupplyShares += feeShares.toUint128();
+            }
+
+            emit EventsLib.AccrueInterest(id, borrowRate, interest, feeShares);
         }
-
-        emit EventsLib.AccrueInterest(id, borrowRate, interest, feeShares);
 
         // Safe "unchecked" cast.
         market[id].lastUpdate = uint128(block.timestamp);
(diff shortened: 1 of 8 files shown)


src/Morpho.sol
```diff
@@ -471,7 +471,7 @@ contract Morpho is IMorphoStaticTyping {
     function _accrueInterest(MarketParams memory marketParams, Id id) internal {
         uint256 elapsed = block.timestamp - market[id].lastUpdate;
 
-        if (elapsed == 0) return;
+        if (elapsed == 0 || marketParams.irm == address(0)) return;
```
QGarchery, line 474: We can avoid touching one slot (the fee/lastUpdate slot) by putting it at the very beginning of the function

QGarchery (approved):

MerlinEgalite (changes requested):
Should be tested i think

MerlinEgalite (approved):

Rubilmax (commented):
Do it in the external library too

MerlinEgalite (dismissed):

MathisGD (changes requested):

src/Morpho.sol
```diff
@@ -469,8 +469,9 @@ contract Morpho is IMorphoStaticTyping {
     /// @dev Accrues interest for the given market `marketParams`.
     /// @dev Assumes that the inputs `marketParams` and `id` match.
     function _accrueInterest(MarketParams memory marketParams, Id id) internal {
-        uint256 elapsed = block.timestamp - market[id].lastUpdate;
+        if (marketParams.irm == address(0)) return;
```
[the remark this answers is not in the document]
Rubilmax: I rebase on #634 and perhaps having #642 rebased on this one will help convince me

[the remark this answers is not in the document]
Rubilmax: Ideally it should, but we currently never emit `lastUpdate` and expect the offchain integrator to hardcode that `lastUpdate = block.timestamp` every time a function that accrues interest is called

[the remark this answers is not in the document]
MathisGD: if we can simplify the logic here we should do it: that's why we need #642

Jean-Grimal (dismissed):

src/Morpho.sol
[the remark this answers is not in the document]
MathisGD: >having https://github.com/morpho-org/morpho-blue/pull/642 rebased on this one will help convince me

ok, will do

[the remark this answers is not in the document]
Rubilmax: So you take care of it in #642?
I think it's better as the code is consistent in its current state, and the spec may be changed in #642

Jean-Grimal (approved):

src/Morpho.sol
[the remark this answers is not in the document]
MathisGD: >I think it's better as the code is consistent in its current state

to me no because now there are some markets where lastUpdated is never updated, and the AccruedInterest event is never emitted. It can be part of the spec for sure, but it complexifies the spec for nothing imo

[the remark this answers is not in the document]
Rubilmax: > now there are some markets where lastUpdated is never updated

makes me think that it saves a lot of gas

> It can be part of the spec for sure, but it complexifies the spec for nothing IMO

So I wouldn't say "for nothing" ^^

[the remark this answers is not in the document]
MathisGD: 2000 something, I think that it's not worth complexifying the spec for that.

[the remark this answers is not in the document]
Rubilmax: Isn't it 5k to SSTORE a cold slot?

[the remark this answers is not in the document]
MathisGD: not for a hot slot

MerlinEgalite (commented):
Well not sure now we should do it

src/Morpho.sol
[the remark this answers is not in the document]
Rubilmax: But how is it hot? it should never been touched ever with this design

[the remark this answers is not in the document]
MerlinEgalite: Because of 

`require(market[id].lastUpdate != 0, ErrorsLib.MARKET_NOT_CREATED);`

MathisGD:
I just reopened [this discussion](https://morpholabs.slack.com/archives/C02N7CZ088N/p1700916873796049), please wait a little bit before merging

src/libraries/periphery/MorphoBalancesLib.sol
```diff
@@ -36,13 +36,12 @@ library MorphoBalancesLib {
...
     {
         Id id = marketParams.id();
-
         Market memory market = morpho.market(id);
 
         uint256 elapsed = block.timestamp - market.lastUpdate;
 
-        // Skipped if elapsed == 0 or if totalBorrowAssets == 0 because interest would be null.
-        if (elapsed != 0 && market.totalBorrowAssets != 0) {
+        // Skipped if elapsed == 0 or totalBorrowAssets == 0 or irm == address(0) because interest would be null.
```
MathisGD, line 43:
```suggestion
        // Skipped if elapsed == 0 or totalBorrowAssets == 0 because interest would be null, or if irm == address(0).
```

MathisGD (approved):

MerlinEgalite (approved):

Jean-Grimal (approved):

MerlinEgalite:
Should we wait for your branch to be merged in it @MathisGD ?

Rubilmax:
I think these 2 PRs are independent

src/Morpho.sol
```diff
@@ -484,25 +484,27 @@ contract Morpho is IMorphoStaticTyping {
...
+                position[id][feeRecipient].supplyShares += feeShares;
+                market[id].totalSupplyShares += feeShares.toUint128();
+            }
 
-        emit EventsLib.AccrueInterest(id, borrowRate, interest, feeShares);
+            emit EventsLib.AccrueInterest(id, borrowRate, interest, feeShares);
+        }
 
         // Safe "unchecked" cast.
         market[id].lastUpdate = uint128(block.timestamp);
```
StErMi, line 510: What's the "meaning" of `lastUpdate`? I'm asking because there's no natspec documentation on the interface, and previously it was updated when there was an interest accrual.

So if I need to give a meaning from the previous code logic, `lastUpdate` represented the last time that interest was accrued or at least triggered (maybe there was no accrual because of market condition or rounding)

With this change instead, `lastUpdate` is always updated even if no interest accrual is triggered, and we know for sure that the market has not changed.

Should the `market[id].lastUpdate = uint128(block.timestamp);` execution be moved inside the `if` branch?

src/Morpho.sol
```diff
@@ -160,7 +160,7 @@ contract Morpho is IMorphoStaticTyping {
         emit EventsLib.CreateMarket(id, marketParams);
 
         // Call to initialize the IRM in case it is stateful.
-        IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
+        if (marketParams.irm != address(0)) IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
```
StErMi, line 163: From my understanding, this PR is to allow the creation of "idle" markets where there's no IRM. Is my understanding correct?

This allows anyone to create a market where no interest will accrue.

- borrowers do not need to pay any interest to suppliers
- suppliers do not accrue and earn any earning from borrowers
- bad debt socialization can't be "neutralized" by interest accrued in the past

Isn't there any better solution to allow the creation of "idle" markets? Because with this modification and the fact that you are not checking during market creation if the `loanToken` and `collateralToken` exists (`!= address(0)` or `address(token).code.length > 0`) it could lead to some unwanted behaviors.

Given that you are shaping MB to allow the MM use case (idle markets), I would suggest thinking about:

1) find the conditions that **only** allow the creation of an idle market (supply `loanToken` without allowing borrowing, only withdraw)
2) does not allow the creation of arbitrary markets that do no make sense for a landing platform

src/Morpho.sol
QGarchery: I think now it refers to the timestamp of the block where there was the last interaction that could change the interest. So mostly the main lending entry points except `supplyCollateral`. I agree that it does not really make the spec simpler in the end, so maybe it's better to use the previous solution of skipping the all function `_accrueInterest` as soon as the irm is 0

StErMi: Do you mean something like `if (elapsed == 0 || marketParams.irm == address(0)) return;`?

src/Morpho.sol
MerlinEgalite: > Is my understanding correct?

Correct

> As I mentioned previously, IMHO, Morpho blue should not be designed and shaped for the needs of externals actors like MetaMorpho. MetaMorpho should be shaped around the basic building block that MB should be.

Indeed, it's 100% aligned with our general philosophy but we think it makes more sense from a product perspective and we won't change it.

src/Morpho.sol
QGarchery: We thought about returning even earlier, before computing `elapsed`, so that we would not have to touch the fee/lastUpdate slot. But this is not such a big optimization actually because this slot is warm anyway (when checking if the market is created). So we thought about your suggestion also.

In the end, because optimizations are not so big, we went for the solution where markets with 0 irm function similarly to regular markets

MathisGD: I think that lastUpdate should always be updated (current situation) because it makes the spec simpler (it's always updated when accrueInterest is called). Also I was in favor (but not big deal) of always emitting the event for the same reason. For me when irm=0 we can say that the interest is accrued, but with rate=0.

src/Morpho.sol
MathisGD: >This allows anyone to create a market where no interest will accrue.
>
>borrowers do not need to pay any interest to suppliers
suppliers do not accrue and earn any earning from borrowers
bad debt socialization can't be "neutralized" by interest accrued in the past

correct, but this is not the expected use-case: for an idle market, one would put address zero as collateral and no borrower would be able to deposit collateral (and so borrow). It's only a way to make these idle depositors pay less gas.

StErMi: I'm confused because this change has been done specifically to allow the creation of "idle market" but it also allows the creation of unwanted (as far as I get) markets.

Why don't you narrow the creation of those market to just the one that makes sense to be used as an idle market (as you said)?

MathisGD: agree that it feels weird, but in Blue you can also created unwanted markets in multiple ways (tokens are not checked, oracles are not checked) so why checking for this thing specifically

StErMi: > agree that it feels weird, but in Blue you can also created unwanted markets in multiple ways (tokens are not checked, oracles are not checked) so why checking for this thing specifically

Because this modification has been made deliberately to allow the creation of an idle market, with the side effects that it allows the creation of non-lending markets. 

The final decision is obviously up to you, I just wanted to point out the side effects and that at least those side effects need to be documented.

PR #642, opened 2023-12-09
refactor `accrueInterest`

MathisGD:
- fixes the weird position of the event emission in `_accrueInterest` (before the write of lastUpdate)
- it removes a early return that are dangerous imo
- it simplifies the spec of the event emission in `_accrueInterest` (always emitted)
- alls irm calls are done through one function
- now blue is irm reentrancy safe

I opened this PR to 1- remove the weird call to the irm, and 2- synchronize the update of “lastUpdate” and the call to the irm
in accrueInterest, elapsed=block.timestamp which is weird but as totalBorrow=0 it’s ok
but now I changed my mind, I think that it’s better to update first “lastUpdate” and have elapsed=0 in accrueInterest


src/Morpho.sol
```diff
@@ -472,41 +468,46 @@ contract Morpho is IMorphoStaticTyping {
     /// @dev Accrues interest for the given market `marketParams`.
     /// @dev Assumes that the inputs `marketParams` and `id` match.
     function _accrueInterest(MarketParams memory marketParams, Id id) internal {
-        uint256 elapsed = block.timestamp - market[id].lastUpdate;
-
-        if (elapsed == 0) return;
-
+        // The IRM is always called, useful for stateful IRMs.
```
MerlinEgalite, line 471: maybe we should why is it useful?

src/Morpho.sol
```diff
@@ -153,14 +153,10 @@ contract Morpho is IMorphoStaticTyping {
         require(isLltvEnabled[marketParams.lltv], ErrorsLib.LLTV_NOT_ENABLED);
         require(market[id].lastUpdate == 0, ErrorsLib.MARKET_ALREADY_CREATED);
 
-        // Safe "unchecked" cast.
-        market[id].lastUpdate = uint128(block.timestamp);
+        _accrueInterest(marketParams, id);
```
MerlinEgalite, line 156: Then `AccrueInterest` is logged before `CreateMarket` which is very weird.  I I think we should do the opposite

src/Morpho.sol
```diff
@@ -153,14 +153,10 @@ contract Morpho is IMorphoStaticTyping {
         require(isLltvEnabled[marketParams.lltv], ErrorsLib.LLTV_NOT_ENABLED);
         require(market[id].lastUpdate == 0, ErrorsLib.MARKET_ALREADY_CREATED);
 
-        // Safe "unchecked" cast.
-        market[id].lastUpdate = uint128(block.timestamp);
```
MerlinEgalite, line 157: why did we removed that?

src/Morpho.sol
```diff
@@ -472,41 +468,46 @@ contract Morpho is IMorphoStaticTyping {
...
-        uint256 elapsed = block.timestamp - market[id].lastUpdate;
-
-        if (elapsed == 0) return;
-
+        // The IRM is always called, useful for stateful IRMs.
         uint256 borrowRate = IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
-        uint256 interest = market[id].totalBorrowAssets.wMulDown(borrowRate.wTaylorCompounded(elapsed));
-        market[id].totalBorrowAssets += interest.toUint128();
-        market[id].totalSupplyAssets += interest.toUint128();
+        // At market creation, elapsed==block.timestamp but totalBorrowAssets==0 so no interest is accrued, as expected.
```
MerlinEgalite, line 473: This is no longer true with your changes no? Since `market[id].lastUpdate = uint128(block.timestamp);` has been removed at market creation

src/Morpho.sol
```diff
@@ -472,41 +468,46 @@ contract Morpho is IMorphoStaticTyping {
...
 
     /// @dev Returns whether the position of `borrower` in the given market `marketParams` is healthy.
     /// @dev Assumes that the inputs `marketParams` and `id` match.
     function _isHealthy(MarketParams memory marketParams, Id id, address borrower) internal view returns (bool) {
-        if (position[id][borrower].borrowShares == 0) return true;
-
-        uint256 collateralPrice = IOracle(marketParams.oracle).price();
-
-        return _isHealthy(marketParams, id, borrower, collateralPrice);
+        if (position[id][borrower].borrowShares == 0) {
```
MerlinEgalite, line 505: should not be part if this PR IMO

MerlinEgalite (changes requested):

src/Morpho.sol
MathisGD: it's done in accrue interest, I find it safer like that

src/Morpho.sol
MathisGD: no I think that it's true

src/Morpho.sol
MathisGD: ok I open on top

src/Morpho.sol
MathisGD: good point, I'm not sure that it's a problem though. It's like that in all functions (I understand that createMarket is different, but i'm not convinced that we should change that)

src/Morpho.sol
MathisGD: I can't find something that fits well.. can you suggest something please, otherwise i'll resolve

src/Morpho.sol
MathisGD: update, I'm re-adding it (the result is the same but finally I think that it's better like that). See what I wrote [here](https://github.com/morpho-org/morpho-blue/pull/642#issue-2034113101). wdyt?

src/Morpho.sol
MathisGD: outdated, see the description of the PR

MathisGD: outdated, see the description of the PR

src/Morpho.sol
Rubilmax: `// Initialize the IRM's state if it is stateful.`

src/Morpho.sol
Rubilmax: We already had the discussion and decided that it was important to have `AccrueInterest` emitted after `CreateMarket`

Rubilmax (changes requested):
The only benefit of this PR to me is that we call the IRM every time the market is interacted with, which can be useful for stateful IRMs
Otherwise, all other changes in the code are drawbacks:
- change in the order in which events are emitted
- `AccrueInterest` events are emitted when interest accrued is zero (so it is useless)
- call to `_accrueInterest` in `createMarket` is more complex
- the IRM costs gas to call

However, it seems to me that this property:
> we call the IRM every time the market is interacted with
is unnecessary, because the stateful IRM can always re-calculate the evolution of the state between blocks, based on the last state that is passed to it every time... So I am currently against this change, because the benefit doesn't cover the cost of drawbacks

MathisGD:
> * change in the order in which events are emitted

if you are only talking about createMarket, this is easily fixable if we agree that it's a problem (I'm still not convinced)

> * `AccrueInterest` events are emitted when interest accrued is zero (so it is useless)

it doesn't cost a lot, and simplifies the spec (accrue interest is always called when the accrueInterest is called).

> * call to `_accrueInterest` in `createMarket` is more complex

what do you mean?

> * the IRM costs gas to call

you mean that we increase the number of calls to the irm, which costs gas right?

> > we call the IRM every time the market is interacted with
> 
> is unnecessary, because the stateful IRM can always re-calculate the evolution of the state between blocks, based on the last state that is passed to it every time... So I am currently against this change, because the benefit doesn't cover the cost of drawbacks

not sure what you mean here. If you want an irm that depends all interactions (including multiple interactions in the same block) how can it recompute this if you don't call it when the state changes.

src/Morpho.sol
MathisGD: yes I remember, but I don't think that it is a real problem as it is like that in all functions. If you want I can change it though.

Jean-Grimal (dismissed):

adhusson:
> now blue is irm reentrancy safe

this is unclear to me?

src/Morpho.sol
MerlinEgalite: > as it is like that in all functions

wdym?

I think we should change it yes since it makes the listening of market creation harder

MathisGD: >wdym?

accrue interest is emitted before the event of the function itself

>I think we should change it yes since it makes the listening of market creation harder

why ?

MerlinEgalite: Because then `AccrueInterest(id, borrowRate, interest, feeShares)` is logged with an unknown `id` and you need to check the second event `CreateMarket(id, marketParams)` to link this `id` to newly created market which is very counter intuitive.

MathisGD: ok, note that you can also rely on accrue interest to check for market creations too

MathisGD: sure I had this in mind but I don't know if this is really a problem (as I never actually use events myself). @julien-devatom thoughts?

src/Morpho.sol
```diff
@@ -475,30 +474,35 @@ contract Morpho is IMorphoStaticTyping {
     /// @dev Accrues interest for the given market `marketParams`.
     /// @dev Assumes that the inputs `marketParams` and `id` match.
     function _accrueInterest(MarketParams memory marketParams, Id id) internal {
-        if (marketParams.irm == address(0)) return;
+        // The IRM is called even when elapsed=0. It can be useful for stateful IRMs.
+        uint256 borrowRate;
+        if (marketParams.irm == address(0)) borrowRate = 0;
+        else borrowRate = IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
```
MerlinEgalite, line 480:
```suggestion
        uint256 borrowRate = (marketParams.irm != address(0) ? IIrm(marketParams.irm).borrowRate(marketParams, market[id]) : 0);
```

MerlinEgalite (approved):

Rubilmax:
Based on https://github.com/morpho-org/metamorpho/pull/269, for a MetaMorpho vault with 10 markets and the AdaptiveCurveIRM, this PR makes a deposit on MM cost on avg 40k more (+5.5% of 715k, min +7k max +117k) which means on avg on a market it costs +4k gas to supply

The tests may not be relevant because 50% of the time, an interaction happens in the same block of another in the same market

MerlinEgalite (approved):

Rubilmax (approved):
I feel like it will be costly at scale because I am so bullish Blue will become the base layer of lending that I expect multiple interactions in a single block
But at worst it leaves room for improvement for V2...

test/forge/integration/CreateMarketIntegrationTest.sol
```diff
@@ -44,17 +44,14 @@ contract CreateMarketIntegrationTest is BaseTest {
 
     function testCreateMarketWithEnabledIrmAndLltv(MarketParams memory marketParamsFuzz) public {
         marketParamsFuzz.lltv = _boundValidLltv(marketParamsFuzz.lltv);
+        marketParamsFuzz.irm = address(irm);
         Id marketParamsFuzzId = marketParamsFuzz.id();
 
         vm.startPrank(OWNER);
         if (!morpho.isIrmEnabled(marketParamsFuzz.irm)) morpho.enableIrm(marketParamsFuzz.irm);
```
Rubilmax, line 51: I believe this is useless now

Jean-Grimal (approved):

test/forge/integration/CreateMarketIntegrationTest.sol
MathisGD:
```suggestion
```

adhusson (approved):

QGarchery (approved):
Last argument against: we just skip the call of the IRM for idle markets, whereas we gained much more before. This is too bad, and I still think the changes in this PR are not worth the potential added gas cost and changing that many lines, but I'm not blocking it

PR #643, opened 2023-12-10
refactor `isHealthy`

MathisGD:
remove a early return, because I think that it is a dangerous design pattern

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -502,11 +502,12 @@ contract Morpho is IMorphoStaticTyping {
     /// @dev Returns whether the position of `borrower` in the given market `marketParams` is healthy.
     /// @dev Assumes that the inputs `marketParams` and `id` match.
     function _isHealthy(MarketParams memory marketParams, Id id, address borrower) internal view returns (bool) {
-        if (position[id][borrower].borrowShares == 0) return true;
-
-        uint256 collateralPrice = IOracle(marketParams.oracle).price();
-
-        return _isHealthy(marketParams, id, borrower, collateralPrice);
+        if (position[id][borrower].borrowShares == 0) {
+            return true;
+        } else {
+            uint256 collateralPrice = IOracle(marketParams.oracle).price();
+            return _isHealthy(marketParams, id, borrower, collateralPrice);
+        }
     }
 
     /// @dev Returns whether the position of `borrower` in the given market `marketParams` with the given


Jean-Grimal (approved):

Rubilmax (commented):
I don't know, I'm not strongly against this pattern but I find it ugly and not less prone to error... Like it's a widely used pattern in a lot of languages and softwares
It's not a point in itself against the fact that it is not prone to error, but rather that it is a widely used and accepted pattern that reviewers are aware of and know how to reason about

While the other is compact and is less readable to me
In any case I know we're debating over a code pattern

Btw this pattern is used not only here so we should harmonize to be consistent?

Rubilmax (approved):

src/Morpho.sol
```diff
@@ -502,11 +502,12 @@ contract Morpho is IMorphoStaticTyping {
     /// @dev Returns whether the position of `borrower` in the given market `marketParams` is healthy.
     /// @dev Assumes that the inputs `marketParams` and `id` match.
     function _isHealthy(MarketParams memory marketParams, Id id, address borrower) internal view returns (bool) {
```
QGarchery, line 504:
```suggestion
    function _isHealthy(MarketParams memory marketParams, Id id, address borrower) internal view returns (bool healthy) {
        healthy = true;
        
        if (position[id][borrower].borrowShares != 0) {
            uint256 collateralPrice = IOracle(marketParams.oracle).price();
            healthy = _isHealthy(marketParams, id, borrower, collateralPrice);
        }
    }
```
What about this ? Would you consider using return variables a case of an "early return" ?

QGarchery (approved):
I'm not sure we should ban the use of early return altogether. Also this function is pretty straightforward, so it wouldn't be a big improvement no matter what we go for.

That being said I'm not blocking the change if you want it

adhusson (approved):
Not a big deal either way

src/Morpho.sol
MathisGD: I also don't like variables that changes value in the middle of the execution :-/

QGarchery (changes requested):
Please wait for the Certora CI run in post-cantina before merging this. We can't see the link in the CI (which is something I reported to the Certora team), and I'm interested in having that result

QGarchery (approved):
Got the result (it's going through)

Rubilmax (approved):

PR #645, opened 2023-12-13
fix(foundry.toml): decrease optimization runs

Rubilmax:
Unfortunately, Etherscan does not support verifying contracts with more than 999999 optimization runs


MerlinEgalite (approved):

MathisGD (approved):
bearish

PR #630, opened 2023-12-05
various fixes

MathisGD:
- [cantina#13](https://cantina.xyz/code/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/598f13e2-bbd5-4fb2-b29f-1335bf426f5e) (comment authorisations are for all markets)
- [cantina#23](https://cantina.xyz/ai/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/672321ef-aa47-492a-bdaa-d5f8247d8b2e) (misleading event natspec)
- [cantina#24](https://cantina.xyz/ai/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/1b99048e-2b2e-4cc3-9727-9d263a2d5a47) (supply collateral)
- [cantina#25](https://cantina.xyz/ai/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/47871fa9-13be-4298-9d00-53735509dbca) (withdraw collateral)
- [cantina#48](https://cantina.xyz/ai/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/8496f70d-95da-41bc-b920-a5c5ca2ef804) (setFee natspec)
- [cantina#90](https://cantina.xyz/ai/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/c735addb-5c2e-48e1-9ad3-dd6619c60223) (front-run repay)
- [cantina#325](https://cantina.xyz/ai/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/39161759-a999-4f1c-bacb-de3f66ba19a6) (bytes.concat)
- [cantina#431](https://cantina.xyz/code/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/eaf8000e-75c3-4a08-b335-b9638009796f) (authorization already set)
- [cantina#503](https://cantina.xyz/code/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/671a881b-38ab-4cfa-8729-372a0c2d9c91) (event emission order) 
- [cantina#670](https://cantina.xyz/code/d86b7f95-e574-4092-8ea2-78dcac2f54f1/findings/ddd3247d-6cfe-4513-8e9b-25c6eade1021?q=status%253Anew%252Cdisputed%252Crejected%252Cduplicate%252Cconfirmed%252Cacknowledged%252Cfixed) (flashloan zero amount)
[...]

PR #652, opened 2023-12-18
fix(authorization): revert already set error

Rubilmax:
- Fixes https://github.com/morpho-org/morpho-blue/pull/630#discussion_r1430371174

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -444,6 +444,7 @@ contract Morpho is IMorphoStaticTyping {
 
     /// @inheritdoc IMorphoBase
     function setAuthorizationWithSig(Authorization memory authorization, Signature calldata signature) external {
+        /// Do not check whether authorization is already set because the nonce increment is a desired side effect.
         require(block.timestamp <= authorization.deadline, ErrorsLib.SIGNATURE_EXPIRED);
         require(authorization.nonce == nonce[authorization.authorizer]++, ErrorsLib.INVALID_NONCE);
 
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -444,10 +444,6 @@ contract Morpho is IMorphoStaticTyping {
 
     /// @inheritdoc IMorphoBase
     function setAuthorizationWithSig(Authorization memory authorization, Signature calldata signature) external {
-        require(
-            authorization.isAuthorized != isAuthorized[authorization.authorizer][authorization.authorized],
-            ErrorsLib.ALREADY_SET
-        );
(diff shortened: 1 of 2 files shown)


adhusson (approved):

QGarchery (approved):
Should also remove the corresponding test, but otherwise LGTM

MerlinEgalite (commented):
we don't delete it for `setAuthorization`? I feel this is useless as well and a bit weird to have 2 different requirements for the same functionality

Rubilmax:
> we don't delete it for `setAuthorization`? I feel this is useless as well and a bit weird to have 2 different requirements for the same functionality

This argument can also be used to defend the introduction of such requirement: why would we require not already set fee but not in setAuthorization* functions?

In the end, I am convinced we were inconsistent, though I am not convinced it is a big deal...

MerlinEgalite:
But now we're being weirdly inconsistent it's even more disturbing to me

Rubilmax:
> But now we're being weirdly inconsistent it's even more disturbing to me

Mb I wasn't clear: I was arguing we shouldn't revert the change at all!

Rubilmax:
Another rationale behind this PR would be to simply revert the changes to the old way, even if it can be considered inconsistent, because it still works and there's no big deal

QGarchery:
> why would we require not already set fee but not in setAuthorization* functions?

For `setAuthorizationWithSig` it's because the function is inherently more complicated becuase signatures are asynchronous by nature. I don't feel like just because it looks disturbing it means that it is incorrect.

You can also look at it another way: we are never in the "already set" case for `setAuthorizationWithSig` because it's at least changing the nonce. If we really want to return `ALREADY_SET`, we could change `INVALID_NONCE` into `ALREADY_SET`

MathisGD (approved):

MathisGD (approved):

Rubilmax (commented):
LGTM

MerlinEgalite (commented):
Let's add a comment no?

Rubilmax:
> Let's add a comment no?

Something like:
`/// Do not check whether authorization is already set because the nonce increment is a desired side-effect.`

MerlinEgalite:
> > Let's add a comment no?
> 
> Something like: `/// Do not check whether authorization is already set because the nonce increment is a desired side-effect.`

LGTM

MerlinEgalite (approved):

MathisGD (approved):

Jean-Grimal (approved):

QGarchery (approved):

PR #653, opened 2023-12-18
Rounding liquidate fix

QGarchery:
See the review of #638 for more details

### DIFF
--- a/src/Morpho.sol
+++ b/src/Morpho.sol
@@ -375,11 +375,8 @@ contract Morpho is IMorphoStaticTyping {
                     market[id].totalBorrowAssets, market[id].totalBorrowShares
                 );
             } else {
-                uint256 repayIncentive = repaidShares.toAssetsDown(
-                    market[id].totalBorrowAssets, market[id].totalBorrowShares
-                ).wMulDown(liquidationIncentiveFactor);
-
-                seizedAssets = repayIncentive.mulDivDown(ORACLE_PRICE_SCALE, collateralPrice);
+                seizedAssets = repaidShares.toAssetsDown(market[id].totalBorrowAssets, market[id].totalBorrowShares)
+                    .wMulDown(liquidationIncentiveFactor).mulDivDown(ORACLE_PRICE_SCALE, collateralPrice);
             }
         }
         uint256 repaidAssets = repaidShares.toAssetsUp(market[id].totalBorrowAssets, market[id].totalBorrowShares);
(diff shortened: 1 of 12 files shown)


adhusson (approved):
Not sure about `repayIncentive` but otherwise yes

src/Morpho.sol
```diff
@@ -371,12 +371,15 @@ contract Morpho is IMorphoStaticTyping {
             if (seizedAssets > 0) {
                 uint256 seizedAssetsQuoted = seizedAssets.mulDivUp(collateralPrice, ORACLE_PRICE_SCALE);
 
-                repaidShares = seizedAssetsQuoted.wDivUp(liquidationIncentiveFactor).toSharesDown(
+                repaidShares = seizedAssetsQuoted.wDivUp(liquidationIncentiveFactor).toSharesUp(
```
MathisGD, line 374: nice catch, do you know if the previous code was dangerous ?

MathisGD, line 374: nice catch, do you know if the previous version was dangerous?

I think that we made the mistake because usually when you repay some assets you want to use `toSharesDown`, but here we are going to recompute the amount repaid after. And indeed we can use that, as as you said it reduces the effective LIF (so it's "in favor of the protocol").

MerlinEgalite:
Why is it a draft? Because of the certora test?

src/Morpho.sol
QGarchery: > do you know if the previous version was dangerous?

I think it's highly unlikely that this would cause issues, because the rounding error is done on shares. This means that to impact the number of assets we would have to be in the case where one more less shares gives one less asset. This should not happen consistently and my guess is that it would be difficult to exploit

src/Morpho.sol
```diff
@@ -371,12 +371,15 @@ contract Morpho is IMorphoStaticTyping {
...
                     market[id].totalBorrowAssets, market[id].totalBorrowShares
                 );
             } else {
-                seizedAssets = repaidShares.toAssetsDown(market[id].totalBorrowAssets, market[id].totalBorrowShares)
-                    .wMulDown(liquidationIncentiveFactor).mulDivDown(ORACLE_PRICE_SCALE, collateralPrice);
+                uint256 repayIncentive = repaidShares.toAssetsDown(
+                    market[id].totalBorrowAssets, market[id].totalBorrowShares
+                ).wMulDown(liquidationIncentiveFactor);
+
+                seizedAssets = repayIncentive.mulDivDown(ORACLE_PRICE_SCALE, collateralPrice);
```
MathisGD, line 382: to rollback

Rubilmax (approved):

MerlinEgalite (approved):

Jean-Grimal (approved):
Clean

PR #656, opened 2023-12-19
test(authorization): front-run address(0) authorization

Rubilmax:


Jean-Grimal (approved):

test/forge/integration/AuthorizationIntegrationTest.sol
```diff
@@ -127,14 +127,23 @@ contract AuthorizationIntegrationTest is BaseTest {
         authorization.nonce = 0;
         authorization.authorizer = vm.addr(privateKey);
 
+        address authorized = _boundAddressNotZero(authorization.authorized);
+        authorization.authorized = address(0);
```
MerlinEgalite, line 131: We should create an independent test I think

Rubilmax: But it would essentially be the same test whereas we can test both at the same time

MerlinEgalite: Then can we move `address authorized = _boundAddressNotZero(authorization.authorized);` lower. Else it seems weird to query it now and set `authorization.authorized` to the address zero.

Rubilmax: i pushed a split because you were right it wasn't testing the same thing

MerlinEgalite (approved):

QGarchery (commented):
LGTM, but to adapt depending on whether we use #652 or not (see https://github.com/morpho-org/morpho-blue/pull/652#issuecomment-1864086371)

PR #659, opened 2023-12-20
[Certora] Equivalent input of liquidate

QGarchery:
Handles https://github.com/morpho-org/morpho-blue/pull/638#issuecomment-1847337038


certora/specs/Health.spec
```diff
@@ -106,3 +107,28 @@ filtered { f -> !f.isView }
 // This invariant ensures that bad debt realization cannot be bypassed.
 invariant alwaysCollateralized(MorphoHarness.Id id, address borrower)
     borrowShares(id, borrower) != 0 => collateral(id, borrower) != 0;
+
+// Checks that passing a seized amount input to liquidate leads to repaid shares S and repaid amount A such that liquidating instead with shares S also repays the amount A.
+rule liquidateEquivalentInputDebtAndInputCollateral(env e, MorphoHarness.MarketParams marketParams, address borrower, uint256 seizedAssets, bytes data) {
+    MorphoHarness.Id id = libId(marketParams);
+
+    // Assume no interest accrual to ease the verification.
+    require lastUpdate(id) == e.block.timestamp;
```
MathisGD, line 116: interesting that this helps

certora/specs/Health.spec
```diff
@@ -106,3 +107,28 @@ filtered { f -> !f.isView }
...
+
+    storage init = lastStorage;
+    uint256 sharesBefore = borrowShares(id, borrower);
+
+    uint256 repaidAssets1;
+    _, repaidAssets1 = liquidate(e, marketParams, borrower, seizedAssets, 0, data);
+    require !priceChanged;
```
MathisGD, line 123: what does this do exactly and why do we need it?

QGarchery: It assumes that the price is the same for the 2 liquidations. It's not really a constraint because in this rule we want to make sure that changing only the input, but in the same context (including sending the tx at the same time), will not change the repaid assets. Usually we don't have to do that because if the storage is the same, then functions (such as `oracle.price()`) do return the same value. But when we use summaries such as `mockPrice()` then we don't have that property for free.

We need it because otherwise a different price would compute a different repaid assets in the end, because of the conversion between collateral and repaid assets

certora/specs/Health.spec
QGarchery: Yes, sometimes even removing computation that does not play a role in the proof can help

certora/specs/Health.spec
MathisGD: `init` is only about the storage state and not about the summaries state is that it ?

MathisGD (approved):

PR #661, opened 2024-01-05
chore: add competition findings

MerlinEgalite:


QGarchery (changes requested):
Same as https://github.com/morpho-org/metamorpho/pull/381#pullrequestreview-1806018877

QGarchery (approved):
> We need a way to expose the findings from precedent audit and security competitions for the upcoming but bounty.
What alternative do you suggest?

Adding it is required to start the bug bounty so I don't see a better alternative: every submitted issue must be shared, and it's easier if it's on the same repo. Plus this is a temporary solution, and the final pdf of the audit will be shared later and will replace this one

PR #664, opened 2024-01-10
fix(ci): fix npm-release action

### LINKED ISSUE #663: Fix NPMjs deployment
https://github.com/morpho-org/morpho-blue/actions/runs/7468066941

Rubilmax:
Fixes #663

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@morpho-org/morpho-blue",
   "description": "Morpho Blue Protocol",
   "license": "BUSL-1.1",
-  "version": "0.1.0",
+  "version": "1.0.0",
   "files": [
     "src",
     "README.md",
--- a/.github/workflows/npm-release.yml
+++ b/.github/workflows/npm-release.yml
@@ -7,9 +7,11 @@ jobs:
   publish-to-npm:
     name: Publish to NPM
     runs-on: ubuntu-latest
+
     environment:
       name: npm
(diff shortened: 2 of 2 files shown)


MerlinEgalite (approved):

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@morpho-org/morpho-blue",
   "description": "Morpho Blue Protocol",
   "license": "BUSL-1.1",
-  "version": "0.1.0",
+  "version": "1.0.0",
```
QGarchery, line 5: It's the tag we used for the release right ? So should we update the release tag as well ?

Rubilmax: I'm not against @MerlinEgalite ?

Jean-Grimal (approved):

package.json
MerlinEgalite: I'd prefer to have released correlated to contract changes but since morpho blue won't change maybe it's not that relevant. I don't mind about both options tbh

MathisGD: the tag is `v1.0.0` what's the problem?

MathisGD (approved):

package.json
QGarchery: > the tag is v1.0.0 what's the problem?

When this PR will be merged, the tag 1.0.0 will still have `0.1.0` in the `version` field of `package.json`, which I find confusing

MathisGD: got it
yes let's update

PR #671, opened 2024-02-26
[Certora] Complete the idToMarketParams invariant

QGarchery:
This PR completes the invariant about `idToMarketParams`. This invariant is splitted into 2 Certora invariants (`hashOfMarketParams` and `marketParamsOfHash`), and states that: `idToMarketParams()` and `id()` are reciprocal functions, under the condition that the corresponding market has been created

### DIFF
--- a/certora/harness/MorphoHarness.sol
+++ b/certora/harness/MorphoHarness.sol
@@ -18,6 +18,10 @@ contract MorphoHarness is Morpho {
         return MAX_FEE;
     }
 
+    function toMarketParams(Id id) external view returns (MarketParams memory) {
+        return idToMarketParams[id];
+    }
+
     function totalSupplyAssets(Id id) external view returns (uint256) {
         return market[id].totalSupplyAssets;
     }
--- a/certora/specs/ConsistentState.spec
+++ b/certora/specs/ConsistentState.spec
@@ -98,10 +98,16 @@ invariant sumBorrowSharesCorrect(MorphoHarness.Id id)
 invariant borrowLessThanSupply(MorphoHarness.Id id)
     totalBorrowAssets(id) <= totalSupplyAssets(id);
 
-// This invariant is useful in the following rule, to link an id back to a market.
-invariant marketInvariant(MorphoHarness.Id id)
+// Check correctness of applying idToMarketParams() to an identifier.
+invariant marketParamsOfHash(MorphoHarness.Id id)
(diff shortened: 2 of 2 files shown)


certora/specs/ConsistentState.spec
```diff
@@ -109,43 +98,52 @@ invariant sumBorrowSharesCorrect(MorphoHarness.Id id)
...
 
+// Check correctness of applying idToMarketParams() to an identifier.
+invariant marketParamsOfHash(MorphoHarness.Id id)
+    isCreated(id) =>
+    libId(toMarketParams(id)) == id;
+
+// Check correctness of applying id() to a market params.
 // This invariant is useful in the following rule, to link an id back to a market.
-invariant marketInvariant(MorphoHarness.MarketParams marketParams)
+invariant hashOfMarketParams(MorphoHarness.MarketParams marketParams)
```
MathisGD, line 108: I don't really understand the naming. Here we take toMarketParams of the hash of the market params, so it should be `marketParamsOfHash` and in the previous invariant we take the hash of the marketParams so it should be hashOfMarketParams no ?

certora/harness/MorphoHarness.sol
```diff
@@ -18,6 +18,10 @@ contract MorphoHarness is Morpho {
         return MAX_FEE;
     }
 
+    function toMarketParams(Id id) external view returns (MarketParams memory) {
```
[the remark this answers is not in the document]
QGarchery: We can't override because `idToMarketParams` is not virtual (and we do not use munging on Blue)

certora/specs/ConsistentState.spec
QGarchery: What you are suggesting is what I had first, and then I thought:
- below, for example on line 122, it gives `marketParamsOfHash(marketParams)` which I would weird
- this invariant (line 108) gives a property about `libId(marketParams)`, which is a hash of a market params

But if you find it more natural then let's change it

QGarchery: Adding a trailing `Of` makes it more natural to read, it gives `marketParamsOfHashOf(marketParams)`

MathisGD: indeed, perfect

MathisGD (approved):

MerlinEgalite (approved):

PR #676, opened 2024-04-03
[Certora] fix exact math timeout

QGarchery:
Todo:
- [x] wait for configuration fixing exact math timeout
- [x] wait for version 7.3.0 on certora-cli

### DIFF
--- a/certora/confs/ExactMath.conf
+++ b/certora/confs/ExactMath.conf
@@ -13,7 +13,6 @@
         "-smt_nonLinearArithmetic true",
         "-solvers [z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5},z3:def{randomSeed=6},z3:def{randomSeed=7},z3:lia2]",
     ],
-    "cache": "none",
     "server": "production",
     "msg": "Morpho Blue Exact Math"
 }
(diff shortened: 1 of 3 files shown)


foundry.toml
```diff
@@ -1,4 +1,5 @@
 [profile.default]
+libs = ["lib"]
```
QGarchery, line 2: This change is to make the versioning independent from usage of the repository. For example, try (without this line) to do:
```bash
rm -rf lib
forge install
git status
```

certora/confs/ExactMath.conf
```diff
@@ -6,10 +6,12 @@
...
     "prover_args": [
-        "-depth 3",
-        "-smt_hashingScheme plaininjectivity",
-        "-mediumTimeout 30",
+        "-depth 5",
+        "-mediumTimeout 5",
         "-timeout 3600",
+        "-adaptiveSolverConfig false",
+        "-smt_nonLinearArithmetic true",
+        "-solvers [z3:def{randomSeed=1},z3:def{randomSeed=2},z3:def{randomSeed=3},z3:def{randomSeed=4},z3:def{randomSeed=5},z3:def{randomSeed=6},z3:def{randomSeed=7},z3:lia2]",
```
MathisGD, line 14: why do you need to specify solvers and why the "random seeds"?

MathisGD (approved):

PR #679, opened 2024-04-25
[Certora] Last timeouts

QGarchery:
This PR addresses timeouts:
- it fixes the timeout on `liquidate` on `onlyAccrueInterestCanIncreaseBorrowRatio`
- it fixes the timeouts on `repay` and `borrow` on `stayHealthy`
- it abandons the `liquidate` case on `stayHealthy` as this seems to be too complicated for the prover, or would require too many changes. The verification has passed in a modified version of the specification for most of the cases though

### DIFF
--- a/certora/specs/RatioMath.spec
+++ b/certora/specs/RatioMath.spec
@@ -59,7 +59,7 @@ rule accrueInterestIncreasesSupplyRatio(env e, MorphoHarness.MarketParams market
     mathint assetsAfter = virtualTotalSupplyAssets(id);
     mathint sharesAfter = virtualTotalSupplyShares(id);
 
-    // Check that the ratio increases: assetsBefore/sharesBefore <= assetsAfter / sharesAfter.
+    // Check that the ratio increases: assetsBefore/sharesBefore <= assetsAfter/sharesAfter.
     assert assetsBefore * sharesAfter <= assetsAfter * sharesBefore;
 }
 
@@ -77,7 +77,7 @@ rule accrueInterestIncreasesBorrowRatio(env e, MorphoHarness.MarketParams market
     mathint assetsAfter = virtualTotalBorrowAssets(id);
     mathint sharesAfter = virtualTotalBorrowShares(id);
 
-    // Check that the ratio increases: assetsBefore/sharesBefore <= assetsAfter / sharesAfter.
+    // Check that the ratio increases: assetsBefore/sharesBefore <= assetsAfter/sharesAfter.
     assert assetsBefore * sharesAfter <= assetsAfter * sharesBefore;
 }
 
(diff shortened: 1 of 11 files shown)


certora/specs/RatioMath.spec
MathisGD, line 1: note that all this could have been called "exchangeRate" (between shares and assets)

certora/specs/RatioMath.spec
```diff
@@ -182,3 +187,25 @@ rule repayDecreasesBorrowRatio(env e, MorphoHarness.MarketParams marketParams, u
     // Check that the ratio decreases: assetsBefore/sharesBefore >= assetsAfter / sharesAfter
     assert assetsBefore * sharesAfter >= assetsAfter * sharesBefore;
 }
+
+rule liquidateDecreasesBorrowRatio(env e, MorphoHarness.MarketParams marketParams, address borrower, uint256 seizedAssets, uint256 repaidShares, bytes data)
```
MathisGD, line 191: It seems that you are not requiring anything more than in `onlyAccrueInterestCanIncreaseBorrowRatio`, why not merging then?

MathisGD: or it is the `require assetsAfter != 1;` ? why does it break without this ? you know ?

certora/specs/RatioMath.spec
QGarchery: done !

certora/specs/RatioMath.spec
QGarchery: Yes it is that require that is needed for the rule to pass, and it corresponds to the case where there is no borrow at all after liquidate (only 1 virtual asset). In case where liquidate removes all borrow, then the exchange rate is reset, which can make so that it increases. But it's fine at this point because there are no users, so it should not impact anybody. This is the same reason why we are checking that the repay all resets the borrow exchange rate [here](https://github.com/morpho-org/morpho-blue/blob/63ead62cd829f4ffb24de5e475f731963451a5a3/certora/specs/ExactMath.spec#L37), instead of checking that it decreases it

MerlinEgalite (approved):

certora/specs/RatioMath.spec
MathisGD: get it
why withdrawing all shares should not do the same thing?

MathisGD (approved):

certora/specs/RatioMath.spec
QGarchery: > why withdrawing all shares should not do the same thing?

The same reason that there is an edge case for repay with a `zeroFloorSub` https://github.com/morpho-org/morpho-blue/blob/7298a03c8b7c47d19a8a5f9220b0f12c49735d2a/src/Morpho.sol#L288 that does not exist for withdraw.
In real scenarios, you cannot resest the supply exchange rate, because you cannot remove all assets. This is coming from the fact that the assets to withdraw are rounded down, whereas the assets to repay are rounded up. So when withdrawing all shares, you always increase the exchange rate

PR #684, opened 2024-07-05
[Certora] Stay healthy liquidate

QGarchery:
Todo:
- [ ] make sure that it goes through
- [x] refactor the rule to align its style with the rest
- [x] wait for it to pass on certora-cli, and switch to certora-cli


certora/specs/Health.spec
```diff
@@ -75,6 +78,44 @@ filtered {
     assert !priceChanged => stillHealthy;
 }
 
+// The general case of liquidate for the stayHealthy rule
+//   - the user checked is not the borrower,
```
MathisGD, line 82: Let's explain why it's sound (or not assume this, and make a branch)

certora/specs/Health.spec
```diff
@@ -75,6 +78,44 @@ filtered {
...
+//   - the market of the liquidation is the market of the user,
```
MathisGD, line 83: it times out without this?

certora/specs/Health.spec
```diff
@@ -75,6 +78,44 @@ filtered {
...
+//   - there is still some borrow on the market after liquidation,
+//   - there was no bad debt realization.
```
MathisGD, line 85: it times out with bad debt realisation?

certora/specs/Health.spec
```diff
@@ -75,6 +78,44 @@ filtered {
...
+rule stayHealthyLiquidate(env e, MorphoHarness.MarketParams marketParams, address borrower, uint256 seizedAssets, bytes data) {
+    MorphoHarness.Id id = libId(marketParams);
+    address user;
+
+    priceChanged = false;
```
MathisGD, line 90: why do we need to do this + `require !priceChanged;`

certora/specs/Health.spec
```diff
@@ -75,6 +78,44 @@ filtered {
...
+
+    liquidate(e, marketParams, borrower, seizedAssets, 0, data);
+    require !priceChanged;
+
+    // Safe require because of the invariant sumBorrowSharesCorrect.
+    require borrowShares(id, user) <= totalBorrowShares(id);
+    // Handle the case where not everything was repaid.
+    require totalBorrowAssets(id) > 0;
+    // Handle the case where the user has some collateral left.
+    require collateral(id, borrower) > 0;
```
MathisGD, line 108: I don't understand why we need to assume this? the account is obviously not healthy if it has some debt but not collateral

certora/specs/Health.spec
QGarchery: If the price changed then every account might become unhealthy as the value of the collateral can be 0. Note that here we are preventing the oracle to be updated in the middle of function calls (for example in the middle of liquidate)

certora/specs/Health.spec
QGarchery: Yes, it times out with bad debt realization. We tried long and hard and gave up in the end

certora/specs/Health.spec
QGarchery: No I think we managed to prove it, but I did not feel like it was worth to add it (distinct storage touched). But we can add it if you think it makes sense

certora/specs/Health.spec
QGarchery: This is the same thing as requiring that there is no bad debt realization, but it should be said more clearly

certora/specs/Health.spec
MathisGD: you think that removing the assumption doesn't help understanding? Otherwise removing assumptions is always better no?

certora/specs/Health.spec
MathisGD: My question was: why do we set priceChanged to false **AND** require !priceChanged; Only the latter isn't enough?

certora/specs/Health.spec
QGarchery: I agree that less assumptions is always better but here we need the assumption to help the prover. The other part (where the market of the liquidation is not the market of the user) has been proved separately, but I did not add it as another rule, because I felt that it was not worth to add that rule. Do you think that we should add it too ?

QGarchery: From our previous conversation actually I remember that you were in favor of adding it, I'll do it and then re-request a review :)

certora/specs/Health.spec
QGarchery: Actually this is verified to be true, with the `assert user != borrower` below. I changed the documentation

certora/specs/Health.spec
QGarchery: You are right, only the latter is enough. Removed in the latest version

certora/specs/StayHealthy.spec
MathisGD, line 1: readme should be adapted

certora/specs/Health.spec
```diff
@@ -2,7 +2,10 @@
 methods {
     function extSloads(bytes32[]) external returns bytes32[] => NONDET DELETE;
 
+    function totalBorrowAssets(MorphoHarness.Id) external returns uint256 envfree;
     function totalBorrowShares(MorphoHarness.Id) external returns uint256 envfree;
+    function virtualTotalBorrowAssets(MorphoHarness.Id) external returns uint256 envfree;
+    function virtualTotalBorrowShares(MorphoHarness.Id) external returns uint256 envfree;
```
MathisGD, line 8:
```suggestion
    function totalBorrowShares(MorphoHarness.Id) external returns uint256 envfree;
```

certora/specs/StayHealthy.spec
```diff
@@ -0,0 +1,121 @@
...
+    // Assume no bad debt realization.
+    require collateral(id, borrower) > 0;
+
+    assert user != borrower;
+    assert debtSharesBefore == borrowShares(id, user);
+    assert debtAssetsBefore >= summaryMulDivUp(debtSharesBefore, virtualTotalBorrowAssets(id), virtualTotalBorrowShares(id));
+
+    bool stillHealthy = isHealthy(marketParams, user);
+    require !priceChanged;
+    assert stillHealthy;
```
MathisGD, line 67: why not? to match `stayHealthy`
```suggestion
    assert !priceChanged => stillHealthy;
```

certora/specs/StayHealthy.spec
```diff
@@ -0,0 +1,121 @@
...
+
+    liquidate(e, liquidationMarketParams, borrower, seizedAssets, 0, data);
+    require !priceChanged;
+
+    // Safe require because of the invariant sumBorrowSharesCorrect.
+    require borrowShares(id, user) <= totalBorrowShares(id);
+
+    bool stillHealthy = isHealthy(marketParams, user);
+    require !priceChanged;
+    assert stillHealthy;
```
MathisGD, line 93:
```suggestion
    assert !priceChanged => stillHealthy;
```

certora/specs/StayHealthy.spec
```diff
@@ -0,0 +1,121 @@
...
+    require !priceChanged;
+
+    // Safe require because of the invariant sumBorrowSharesCorrect.
+    require borrowShares(id, user) <= totalBorrowShares(id);
+    // Assume that there is still some borrow on the market after liquidation.
+    require totalBorrowAssets(id) == 0;
+
+    bool stillHealthy = isHealthy(marketParams, user);
+    require !priceChanged;
+    assert stillHealthy;
```
MathisGD, line 120:
```suggestion
    assert !priceChanged => stillHealthy;
```

certora/specs/StayHealthy.spec
```diff
@@ -0,0 +1,121 @@
...
+    require marketParams.lltv < 10^18;
+    // Assumption to ensure that no interest is accumulated.
+    require lastUpdate(id) == e.block.timestamp;
+
+    liquidate(e, marketParams, borrower, seizedAssets, 0, data);
+    require !priceChanged;
+
+    // Safe require because of the invariant sumBorrowSharesCorrect.
+    require borrowShares(id, user) <= totalBorrowShares(id);
+    // Assume that there is still some borrow on the market after liquidation.
```
MathisGD, line 115:
```suggestion
    // Assume that there is no remaining borrow on the market after liquidation.
```

certora/specs/StayHealthy.spec
```diff
@@ -0,0 +1,121 @@
...
+
+rule stayHealthyLiquidateLastBorrow(env e, MorphoHarness.MarketParams marketParams, address borrower, uint256 seizedAssets, bytes data) {
+    MorphoHarness.Id id = libId(marketParams);
+    address user;
+
+    // Assume the invariant initially.
+    require isHealthy(marketParams, user);
+
+    uint256 debtSharesBefore = borrowShares(id, user);
+    uint256 debtAssetsBefore = summaryMulDivUp(debtSharesBefore, virtualTotalBorrowAssets(id), virtualTotalBorrowShares(id));
```
MathisGD, line 104:
```suggestion
```

certora/specs/StayHealthy.spec
```diff
@@ -0,0 +1,121 @@
...
+    require !priceChanged;
+
+    // Safe require because of the invariant sumBorrowSharesCorrect.
+    require borrowShares(id, user) <= totalBorrowShares(id);
+    // Assume that there is still some borrow on the market after liquidation.
+    require totalBorrowAssets(id) > 0;
+    // Assume no bad debt realization.
+    require collateral(id, borrower) > 0;
+
+    assert user != borrower;
```
MathisGD, line 61: technically, this should be verifiable too (but might timeout)
```suggestion
```

certora/specs/StayHealthy.spec
```diff
@@ -0,0 +1,121 @@
...
+    assert debtSharesBefore == borrowShares(id, user);
+    assert debtAssetsBefore >= summaryMulDivUp(debtSharesBefore, virtualTotalBorrowAssets(id), virtualTotalBorrowShares(id));
```
MathisGD, line 63: why do we need to assert this before checking stillHealthy? Helps the prover?

certora/specs/Health.spec
QGarchery: Those are actually needed in `StayHealthy.spec` now

certora/specs/StayHealthy.spec
QGarchery: This is to help the prover actually, to consider only one path: the one where price did not change at all. I tried removing and simplifying some of those and it times out

certora/specs/StayHealthy.spec
QGarchery: same

certora/specs/StayHealthy.spec
QGarchery: same

certora/specs/StayHealthy.spec
QGarchery: It is actually verified ! This is an assert, not a require

certora/specs/StayHealthy.spec
QGarchery: Yes, this is a hint to the prover

certora/specs/StayHealthy.spec
MathisGD: oh yes sure!

MathisGD (approved):

certora/specs/StayHealthy.spec
```diff
@@ -0,0 +1,118 @@
...
+    require liquidationMarketParams != marketParams;
+
+    liquidate(e, liquidationMarketParams, borrower, seizedAssets, 0, data);
+    require !priceChanged;
+
+    // Safe require because of the invariant sumBorrowSharesCorrect.
+    require borrowShares(id, user) <= totalBorrowShares(id);
+
+    bool stillHealthy = isHealthy(marketParams, user);
+    require !priceChanged;
```
colin-morpho, line 91: Do we want a require or an assert ?

colin-morpho: Why not use an implication

```suggestion
    !priceChanged => property;
```

colin-morpho: Just saw the other comments resolving this thread.

certora/specs/StayHealthy.spec
colin-morpho: Certor being certora.

certora/specs/StayHealthy.spec
colin-morpho, line 1: `StayHealthy` sounds like liveness to me. That is, something good will always happen. Here it goes like that, with given assumptions the transaction result in healthy positions (this is the good thing happening). What about `PositionLiveness.spec` ?? Feel free to dismiss this comment.

colin-morpho (approved):
Minor comment, can be dismissed.

certora/specs/StayHealthy.spec
MathisGD: Not to me: if some interactions could make people's position unhealthy, we wouldn't say there is a liveness issue right? Positions not getting unhealthy is well a safety property.

PR #689, opened 2024-08-14
[Certora] Improve assets accounting specification

QGarchery:
Improve assets accounting specification so that we don't have to assume that supply, borrow, and supplyCollateral start with an empty position

### DIFF
--- a/certora/specs/AssetsAccounting.spec
+++ b/certora/specs/AssetsAccounting.spec
@@ -5,6 +5,8 @@ methods {
     function supplyShares(MorphoHarness.Id, address) external returns uint256 envfree;
     function borrowShares(MorphoHarness.Id, address) external returns uint256 envfree;
     function collateral(MorphoHarness.Id, address) external returns uint256 envfree;
+    function totalSupplyShares(MorphoHarness.Id) external returns uint256 envfree;
+    function totalBorrowShares(MorphoHarness.Id) external returns uint256 envfree;
     function virtualTotalSupplyAssets(MorphoHarness.Id) external returns uint256 envfree;
     function virtualTotalSupplyShares(MorphoHarness.Id) external returns uint256 envfree;
     function virtualTotalBorrowAssets(MorphoHarness.Id) external returns uint256 envfree;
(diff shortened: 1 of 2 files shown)


certora/specs/AssetsAccounting.spec
```diff
@@ -64,22 +68,24 @@ rule withdrawAssetsAccounting(env e, MorphoHarness.MarketParams marketParams, ui
...
+// Check that the increase of owed assets are greater than the borrowed assets.
 rule borrowAssetsAccounting(env e, MorphoHarness.MarketParams marketParams, uint256 shares, address onBehalf, address receiver) {
     MorphoHarness.Id id = libId(marketParams);
 
     // Assume no interest as it would increase the total borrowed assets.
     require lastUpdate(id) == e.block.timestamp;
-    // Assume no outstanding debt to begin with.
-    require borrowShares(id, onBehalf) == 0;
+    // Safe require because of the sumBorrowSharesCorrect invariant.
+    require borrowShares(id, onBehalf) <= totalBorrowShares(id);
```
MathisGD, line 78: not sure why you need to do this, the implem would revert when doing the addition no?

certora/specs/AssetsAccounting.spec
```diff
@@ -32,21 +34,23 @@ function expectedBorrowAssets(MorphoHarness.Id id, address user) returns uint256
...
+// Check that the assets supplied are greater than the increase in owned assets.
 rule supplyAssetsAccounting(env e, MorphoHarness.MarketParams marketParams, uint256 assets, uint256 shares, address onBehalf, bytes data) {
     MorphoHarness.Id id = libId(marketParams);
 
     // Assume no interest as it would increase the total supply assets.
     require lastUpdate(id) == e.block.timestamp;
-    // Assume no supply position to begin with.
-    require supplyShares(id, onBehalf) == 0;
+    // Safe require because of the sumSupplySharesCorrect invariant.
+    require supplyShares(id, onBehalf) <= totalSupplyShares(id);
```
MathisGD, line 44: same question

certora/specs/AssetsAccounting.spec
QGarchery: Which addition are you talking about ?
We need it here because otherwise you could have much more shares than the total, so supplying artificially increases the assets you are owned (in other words the accounting is broken). But this is a state that can never be reached

MathisGD (approved):

MerlinEgalite (approved):

PR #690, opened 2024-08-18
docs: fix spelling issues

nnsW3:
Hello
I found several minor spelling issues.
Br, Elias.

### DIFF
--- a/certora/specs/ExactMath.spec
+++ b/certora/specs/ExactMath.spec
@@ -58,7 +58,7 @@ rule repayAllResetsBorrowExchangeRate(env e, MorphoHarness.MarketParams marketPa
     // There are at least as many shares as virtual shares.
 }
 
-// There should be no profit from supply followed immediately by withdraw.
+// There should be no profit from supply followed immediately by withdrawal.
 rule supplyWithdraw() {
     MorphoHarness.MarketParams marketParams;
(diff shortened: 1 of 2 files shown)


certora/specs/ExactMath.spec
```diff
@@ -58,7 +58,7 @@ rule repayAllResetsBorrowExchangeRate(env e, MorphoHarness.MarketParams marketPa
     // There are at least as many shares as virtual shares.
 }
 
-// There should be no profit from supply followed immediately by withdraw.
```
MathisGD, line 61: "withdraw" is coherent here, as we say "supply"

src/interfaces/IMorphoCallbacks.sol
```diff
@@ -46,7 +46,7 @@ interface IMorphoSupplyCollateralCallback {
 interface IMorphoFlashLoanCallback {
     /// @notice Callback called when a flash loan occurs.
     /// @dev The callback is called only if data is not empty.
-    /// @param assets The amount of assets that was flash loaned.
```
MathisGD, line 49: I think that "was" is ok because it relates to "the amount"

MathisGD (commented):
Thanks for your contribution! But I don't think there are issue, feel free to reopen if you don't think so.

PR #692, opened 2024-09-27
test: update halmos tests for new version

daejunpark:
this updates halmos tests for the new version.

changes:
- replaced --symbolic-storage with enableSymbolicStorage() cheatcode
- explicitly updated the block number and timestamp
- fixed and improved calldata generation for flashLoan()

### DIFF
--- a/test/halmos/HalmosTest.sol
+++ b/test/halmos/HalmosTest.sol
@@ -57,6 +57,7 @@ contract HalmosTest is SymTest, Test {
         flashBorrower = new FlashBorrowerMock(morpho);
 
         // Enable symbolic storage
+        svm.enableSymbolicStorage(address(this));
         svm.enableSymbolicStorage(address(morpho));
         svm.enableSymbolicStorage(address(loanToken));
         svm.enableSymbolicStorage(address(collateralToken));
--- a/test/halmos/HalmosTest.sol
+++ b/test/halmos/HalmosTest.sol
@@ -13,7 +13,7 @@ import "../../src/Morpho.sol";
 import "../../src/libraries/ConstantsLib.sol";
 import {MorphoLib} from "../../src/libraries/periphery/MorphoLib.sol";
 
-/// @custom:halmos --symbolic-storage --solver-timeout-assertion 0
+/// @custom:halmos --solver-timeout-assertion 0
 contract HalmosTest is SymTest, Test {
     using MorphoLib for IMorpho;
     using MarketParamsLib for MarketParams;
@@ -47,6 +47,17 @@ contract HalmosTest is SymTest, Test {
         morpho.enableLltv(lltv);
         morpho.createMarket(marketParams);
         vm.stopPrank();
+
+        // Enable symbolic storage
+        svm.enableSymbolicStorage(address(morpho));
+        svm.enableSymbolicStorage(address(loanToken));
+        svm.enableSymbolicStorage(address(collateralToken));
+        svm.enableSymbolicStorage(address(oracle));
+        svm.enableSymbolicStorage(address(irm));
+
+        // Set symbolic block number and timestamp
+        vm.roll(svm.createUint(64, "block.number"));
+        vm.warp(svm.createUint(64, "block.timestamp"));
     }
 
     // Call Morpho, assuming interacting with only the defined market for performance reasons.
--- a/test/halmos/HalmosTest.sol
+++ b/test/halmos/HalmosTest.sol
@@ -8,6 +8,7 @@ import {IMorpho} from "../../src/interfaces/IMorpho.sol";
 import {IrmMock} from "../../src/mocks/IrmMock.sol";
 import {ERC20Mock} from "../../src/mocks/ERC20Mock.sol";
 import {OracleMock} from "../../src/mocks/OracleMock.sol";
+import {FlashBorrowerMock} from "../../src/mocks/FlashBorrowerMock.sol";
 
 import "../../src/Morpho.sol";
 import "../../src/libraries/ConstantsLib.sol";
@@ -29,6 +30,9 @@ contract HalmosTest is SymTest, Test {
 
     MarketParams internal marketParams;
 
+    ERC20Mock internal otherToken;
+    FlashBorrowerMock internal flashBorrower;
+
     function setUp() public virtual {
         owner = svm.createAddress("owner");
         morpho = IMorpho(address(new Morpho(owner)));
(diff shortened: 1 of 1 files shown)


test/halmos/HalmosTest.sol
```diff
@@ -8,12 +8,13 @@ import {IMorpho} from "../../src/interfaces/IMorpho.sol";
...
 import {ERC20Mock} from "../../src/mocks/ERC20Mock.sol";
 import {OracleMock} from "../../src/mocks/OracleMock.sol";
+import {FlashBorrowerMock} from "../../src/mocks/FlashBorrowerMock.sol";
 
 import "../../src/Morpho.sol";
 import "../../src/libraries/ConstantsLib.sol";
 import {MorphoLib} from "../../src/libraries/periphery/MorphoLib.sol";
 
-/// @custom:halmos --symbolic-storage --solver-timeout-assertion 0
+/// @custom:halmos --solver-timeout-assertion 0
```
daejunpark, line 17: `--symbolic-storage` is no longer available in v2.

test/halmos/HalmosTest.sol
```diff
@@ -47,6 +51,23 @@ contract HalmosTest is SymTest, Test {
...
+        flashBorrower = new FlashBorrowerMock(morpho);
+
+        // Enable symbolic storage
+        svm.enableSymbolicStorage(address(morpho));
+        svm.enableSymbolicStorage(address(loanToken));
+        svm.enableSymbolicStorage(address(collateralToken));
+        svm.enableSymbolicStorage(address(oracle));
+        svm.enableSymbolicStorage(address(irm));
+        svm.enableSymbolicStorage(address(otherToken));
+        svm.enableSymbolicStorage(address(flashBorrower));
```
daejunpark, line 66: svm.enableSymbolicStorage() is now used in place of --symbolic-storage.

test/halmos/HalmosTest.sol
```diff
@@ -47,6 +51,23 @@ contract HalmosTest is SymTest, Test {
...
+
+        // Set symbolic block number and timestamp
+        vm.roll(svm.createUint(64, "block.number"));
+        vm.warp(svm.createUint(64, "block.timestamp"));
```
daejunpark, line 70: the block number and timestamp are no longer symbolic, but initialized as they are in Foundry. they need to be explicitly updated as in Foundry tests.

test/halmos/HalmosTest.sol
```diff
@@ -74,17 +95,9 @@ contract HalmosTest is SymTest, Test {
...
-            address token;
-            if (rand == 0) {
-                token = address(loanToken);
-            } else if (rand == 1) {
-                token = address(collateralToken);
-            } else {
-                ERC20Mock otherToken = new ERC20Mock();
-                token = address(otherToken);
-            }
-            args = abi.encode(marketParams, token, assets, emptyData);
```
daejunpark, line 87: `flashLoan()` doesn't take `marketParams` as input.

test/halmos/HalmosTest.sol
```diff
@@ -74,17 +95,9 @@ contract HalmosTest is SymTest, Test {
...
-            uint256 rand = svm.createUint256("rand");
-            address token;
-            if (rand == 0) {
-                token = address(loanToken);
-            } else if (rand == 1) {
-                token = address(collateralToken);
-            } else {
-                ERC20Mock otherToken = new ERC20Mock();
-                token = address(otherToken);
-            }
```
daejunpark, line 86: explicit case splitting is no longer needed. when the call target address is symbolic, all available addresses are automatically considered.

test/halmos/HalmosTest.sol
```diff
@@ -74,17 +95,9 @@ contract HalmosTest is SymTest, Test {
...
-            args = abi.encode(marketParams, token, assets, emptyData);
+            address token = svm.createAddress("token");
+            bytes memory _data = svm.createBytes(1024, "_data");
+            args = abi.encode(token, assets, _data);
```
daejunpark, line 100: non-empty `_data` is needed for FlashBorrowerMock.onMorphoFlashLoan().

test/halmos/HalmosTest.sol
```diff
@@ -47,6 +51,23 @@ contract HalmosTest is SymTest, Test {
         morpho.enableLltv(lltv);
         morpho.createMarket(marketParams);
         vm.stopPrank();
+
+        // for flashLoan
+        otherToken = new ERC20Mock();
+        flashBorrower = new FlashBorrowerMock(morpho);
```
daejunpark, line 57: a contract that implements `onMorphoFlashLoan()` is needed to execute flashLoan(). other mock contracts can also be used.

test/halmos/HalmosTest.sol
QGarchery: I see that it expects `(address assets, bytes data)` as argument. So what happens if it is empty ? Does it mean that the call revert, and so Halmos considers it as a revert path in the verification ?

test/halmos/HalmosTest.sol
QGarchery: Good to know ! Thanks !

test/halmos/HalmosTest.sol
QGarchery: Right, it makes sense now to enable for all possible contracts !

test/halmos/HalmosTest.sol
QGarchery: What happens if we put those definitions in the constructor of the current contract ? Would it still be considered as valid targets for symbolic addresses ?

daejunpark: do you mean adding onMorphoFlashLoan() in the test contract, right? yes, it will work. any addresses including the test contract will be considered as a target for symbolic addresses.

test/halmos/HalmosTest.sol
daejunpark: you may also want to include `this` address, especially if you add some callback functions in the test contract:
```
svm.enableSymbolicStorage(address(this));
```

test/halmos/HalmosTest.sol
daejunpark: yes. if data is empty, then onMorphoFlashLoan will revert due to `abi.decode(data, (address))`. as a result, flashLoan() will always revert. since we have `vm.assume(success);`, all flashLoan() paths will be silently ignored. essentially, this causes flashLoan() to be omitted from the verification.

test/halmos/HalmosTest.sol
daejunpark: added in https://github.com/morpho-org/morpho-blue/pull/692/commits/a80998eba45966881cb4ab645395b38f710f429f

test/halmos/HalmosTest.sol
QGarchery: > do you mean adding onMorphoFlashLoan() in the test contract?

I was more thinking about having those 2 lines inside `constructor() { ... }` instead of inside `function setUp() { ... }` where it is currently. I'm asking this because the line `ERC20Mock internal otherToken;` could be changed to `ERC20Mock internal otherToken = new ERC20Mock();` instead

QGarchery (approved):

test/halmos/HalmosTest.sol
daejunpark: I see. Yes, you could do that for otherToken, but not for flashBorrower as it needs `morpho` as input.

MathisGD (approved):

MathisGD:
Thanks for your update! @daejunpark

PR #704, opened 2024-11-22
[Certora] Liquidate buffer

QGarchery:


.github/workflows/certora.yml
yoav-el-certora, line 1: Please change line 53 to this-

`run: certoraRun certora/confs/${{ matrix.conf }}.conf --wait_for_results`

QGarchery:
replaced by #708

PR #708, opened 2024-12-09
[Certora] Liquidate buffer, with executable code

QGarchery:
This verification is meant to show that if liquidation happen with LTV < 1 / LIF (note that 1 / LIF > LLTV by design), then this liquidation is healthy in the following sense: the position's health factor increases. This means that liquidation happening in the buffer [LLTV, 1/LIF] always lead to healthy positions, if liquidating multiple times, and don't lead to bad debt


QGarchery:
It seems very difficult to make this verification go though. [Here](https://prover.certora.com/output/40748/271b6a8f4870491bb0e7e686482de10f?anonymousKey=2f3f41d0abac98cfb03251d59e7dd95d60d34c8f) is an example of such run, but deviating a little bit from it seems to cause a timeout. Also this run requires some heavy modifications:
- passing the LIF to the liquidate function
- munging to not update the value of repaidAssets
- munging to comment out _accrueInterest (even though it should be skipped given the assumptions)
- munging to require that no bad debt is realized (even though it should be provable by the assumptions)

For these reasons, it seems prefereable to do #710

QGarchery:
A new trick (see [9ee0ccf](https://github.com/morpho-org/morpho-blue/pull/708/commits/9ee0ccfb1e993d87ee2508c617fdf06da499c629)) is making the verification much faster for the prover. So now this is considered over #710

colin-morpho (approved):
Blazing fast, hard to believe it could ever timeout.

colin-morpho (approved):

certora/specs/LiquidateBuffer.spec
```diff
@@ -0,0 +1,71 @@
...
+    require data.length == 0;
+
+    MorphoHarness.Id id = Util.libId(marketParams);
+
+    // We place ourselves at the last block for getting the following variables.
+    require lastUpdate(id) == e.block.timestamp;
+
+    uint256 borrowerShares = borrowShares(id, borrower);
+    // Safe require because of the sumBorrowSharesCorrect invariant.
+    require borrowerShares <= totalBorrowShares(id);
```
MathisGD, line 41: could we requireInvariant here?

certora/specs/LiquidateBuffer.spec
```diff
@@ -0,0 +1,71 @@
...
+
+    // Hint for the prover to show that there is no bad debt realization.
+    assert newBorrowerCollateral != 0;
+    // Hint for the prover about the ratio used to close the position.
+    assert repaidShares * borrowerCollateral >= seizedAssets * borrowerShares;
+    // Prove that the ratio of shares of debt over collateral is smaller after the liquidation.
+    assert borrowerShares * newBorrowerCollateral >= newBorrowerShares * borrowerCollateral;
+    // Prove that the value of borrow shares is smaller after the liquidation.
+    // Note that this is only shown for the case where there are still borrow positions on the markets.
+    assert totalBorrowAssets(id) > 0 => newVirtualTotalShares * virtualTotalAssets >= newVirtualTotalAssets * virtualTotalShares;
```
MathisGD, line 70: why do we prove this? for other positions?

MathisGD (approved):

certora/specs/LiquidateBuffer.spec
QGarchery: We prove 2 things:
- that shares/collateral improves / is reduced
- that the values of shares (to convert them into assets) also improves / is reduced

This way we can show that, using math (non rounding operations), the LTV improves, where LTV is assets/collateral.
[Reference slack thread](https://morpholabs.slack.com/archives/C02N7CZ088N/p1733500270289979?thread_ts=1733500250.758919&cid=C02N7CZ088N)

certora/specs/LiquidateBuffer.spec
QGarchery: We could not do it directly here, because the invariant is not saying this explicitly. Rather it is showing that totalBorrowShares is equal to the ghost value that is updated each time a borrowShares gets changed. We could use some of the same tricks that we used in morpho-token, with partial sums, to make that explicit. But it would require some significant change to the spec, so I'm not sure if it is worth it at this point

PR #715, opened 2025-01-14
fix: rename harness to helpers

### LINKED ISSUE #701: [Certora] `harness` directory name
The directory [harness](certora/harness) should be renamed to `helpers` as it doesn't just contain a harness but also miscellaneous utilities.

mahmudsudo:
closes #701


QGarchery (approved):
Good to go, as verification goes through when I run it locally (this is an issue with CERTORAKEY secret here):
Screenshot from 2025-01-15 10-32-37
Notice how there is one job that is failing, but it's a timeout, and we are already aware of it

MathisGD (approved):
Thanks for your contribution!

PR #717, opened 2025-01-22
remove unintentional single quote

gruns:


MathisGD (approved):
Actually, it's not a mistake (it means the folders of the audits) and we have it [everywhere](https://github.com/morpho-org/metamorpho?tab=readme-ov-file#audits). Though I prefer your version. Thanks!

gruns:
ha. i should have grepped for all occurrences 🙂 

either way, glad i could 

happy to open more PRs here if helpful. will run grep+replace for the equivalent of `s/audits'/audits/g`

MathisGD:
Let's keep repos harmonised, in the end

PR #725, opened 2025-04-04
docs: document loan asset donation

MathisGD:


src/interfaces/IMorpho.sol
```diff
@@ -112,6 +112,8 @@ interface IMorphoBase {
     /// on `transfer` and `transferFrom`. In particular, tokens with fees on transfer are not supported.
     /// - The IRM should not re-enter Morpho.
     /// - The oracle should return a price with the correct scaling.
+    /// - The oracle should not be able to go up instantly by a multiplier of more than 1/(LLTV*LIF). In particular, if
+    /// the loan asset is a vault, the oracle should not price its shares using the AUM.
```
adhusson, line 116:
```suggestion
    /// the loan asset is a MetaMorpho vault, the oracle should not price its shares using the AUM.
```

or

```suggestion
    /// the loan asset is a vault that can receive donations, the oracle should not price its shares using the AUM.
```

4626 vaults in general could be donation resistant and exclude donations from AUM in the sense of `totalAssets`.

adhusson (dismissed):

src/interfaces/IMorpho.sol
```diff
@@ -112,6 +112,8 @@ interface IMorphoBase {
     /// on `transfer` and `transferFrom`. In particular, tokens with fees on transfer are not supported.
     /// - The IRM should not re-enter Morpho.
     /// - The oracle should return a price with the correct scaling.
+    /// - The oracle should not be able to go up instantly by a multiplier of more than 1/(LLTV*LIF). In particular, if
```
QGarchery, line 115:
```suggestion
    /// - The oracle should not be able to go down instantly by a multiplier of less than LLTV*LIF. In particular, if
```
This is the opposite because the oracle in Morpho Blue is pricing the collateral quoted in the loan token, but the problematic case is when the loan token value goes up

MathisGD: you are right!

QGarchery (approved):

src/interfaces/IMorpho.sol
```diff
@@ -112,6 +112,8 @@ interface IMorphoBase {
     /// on `transfer` and `transferFrom`. In particular, tokens with fees on transfer are not supported.
     /// - The IRM should not re-enter Morpho.
     /// - The oracle should return a price with the correct scaling.
+    /// - The oracle should not be able to go down instantly by a multiplier of less than LLTV*LIF. In particular, if
```
adhusson, line 115: "go down by a multiplier" is less clear now (although correct!), due to the possible confusion between decreasing by a factor vs. being multiplied by 1/factor.

```suggestion
    /// - The oracle should not change instantly such that the new price is less than the old price times LLTV*LIF. In particular, if
```

adhusson (approved):

QGarchery (approved):

PR #726, opened 2025-04-22
Add specs and confs for foundry tests under `test/forge/libraries`

naftali-g:
Additionally add a script that triggers all of them.
Note - for a few tests in `SafeTransferLibTest` the fact that we start from an arbitrary state necessitated modifying those test functions a little to make them more robust (and actually improve their coverage).


naftali-g:
This is not yet ready for review, so I'll close it for now

PR #733, opened 2025-07-12
invariant testing config

MathisGD:


QGarchery (approved):

QGarchery:
> Could you give a general idea of why?

I think it's just that you don't expect to have many bugs that are discoverable after > 32 interactions, but this is purely an estimate that we can play with. I fully agree with the direction though, I think that it's more likely to find a bug with less than 4 interactions but with well chosen values for the arguments (which is more likely to be discovered with higher number of runs)

PR #734, opened 2025-08-13
[Certora] Update to prover v8

QGarchery:
See https://docs.certora.com/en/latest/docs/prover/changelog/prover_changelog.html#august-11-2025

Also the change in Liveness is because they also fixed the fact that CVL functions can revert.


MathisGD (approved):

colin-morpho (approved):
LGTM

PR #741, opened 2025-11-28
Halmos Tests with Symbolic IRM

bhargavbh:
Added a symbolic IRM contract which returns a non-deterministic rate for the borrowRate() and borrowRateView() functions. The Morpho contract now calls a symbolic IRM instead of Mock IRM. Halmos tests pass with slightly higher path exploration.

### DIFF
--- a/test/halmos/HalmosTest.sol
+++ b/test/halmos/HalmosTest.sol
@@ -19,8 +19,6 @@ import {MathLib} from "../../src/libraries/MathLib.sol";
 
 // A symbolic IRM for Halmos Tests.
 contract IrmSymbolic is IIrm, SymTest, Test {
-    using MathLib for uint128;
-
     function borrowRateView(MarketParams memory, Market memory) public pure returns (uint256) {
         // Returns a symbolic borrow rate.
         return 50;


bhargavbh:
the linter fails because i cast a Uint256 into Uint128. Halmos only has the symbolic version for [Uint256](https://github.com/a16z/halmos-cheatcodes/blob/main/src/SVM.sol) but not Uint128. Hence i cast it to Uint128. For a halmos test, this is perfectly reasonable and sound since i am casting down and all the symbolic values of Uint128 are covered. Is there a better way around? @QGarchery

test/halmos/HalmosTest.sol
MathisGD: you need to run `forge fmt` (see [this](https://github.com/morpho-org/morpho-blue/actions/runs/19757546350/job/56611875998?pr=741))

bhargavbh:
interestingly, all tests pass with symbolic initialised market except `check_borrowLessThanSupply()` which is the only property reasoning about `totalSupplyAssets` and `totalBorrowAssets`. Verifying it generates 46 constraints but yices timesout on solving the last constraint.

bhargavbh:
closing this PR as the properties proved with Halmos now have certora proofs, check https://github.com/morpho-org/morpho-blue/pull/747.

PR #743, opened 2025-12-04
chore: clean repo

MathisGD:


.github/workflows/formatting.yml
QGarchery: can probably also remove the Hardhat and Node modules section in the .gitignore

QGarchery (approved):

peyha (commented):
~~why do we keep the Makefile ?~~ just saw quentin's remark above

PR #760, opened 2026-02-11
docs: rename misleading prevBorrowRate param in AccrueInterest event

aryansalam:
## Summary
- Rename `prevBorrowRate` to `borrowRate` in the `AccrueInterest` event definition and update NatSpec comment

## Context

In `Morpho.sol`, the `_accrueInterest` function fetches the current borrow rate from the IRM and passes it directly to the `AccrueInterest` event emission ([`Morpho.sol:503`](https://github.com/morpho-org/morpho-blue/blob/main/src/Morpho.sol#L503)):

```solidity
uint256 borrowRate = IIrm(marketParams.irm).borrowRate(marketParams, market[id]);
// ...
emit EventsLib.AccrueInterest(id, borrowRate, interest, feeShares);
```

However, the event parameter was named `prevBorrowRate` with documentation stating "The previous borrow rate." This is misleading — the value is the **current** borrow rate used for the accrual, not a historical value.

## Changes
- Renamed event parameter from `prevBorrowRate` → `borrowRate`
- Updated NatSpec from "The previous borrow rate." → "The borrow rate per second (scaled by WAD)."

## Impact
None. Event parameter names are not part of the ABI selector (`keccak256("AccrueInterest(bytes32,uint256,uint256,uint256)")`), so this is purely a documentation/readability fix with zero downstream breakage.

## Test plan
- [x] All 145 existing tests pass with no modifications

### DIFF
--- a/src/libraries/EventsLib.sol
+++ b/src/libraries/EventsLib.sol
@@ -143,8 +143,8 @@ library EventsLib {
 
     /// @notice Emitted when accruing interest.
     /// @param id The market id.
-    /// @param prevBorrowRate The previous borrow rate.
+    /// @param borrowRate The borrow rate per second (scaled by WAD).
     /// @param interest The amount of interest accrued.
     /// @param feeShares The amount of shares minted as fee.
-    event AccrueInterest(Id indexed id, uint256 prevBorrowRate, uint256 interest, uint256 feeShares);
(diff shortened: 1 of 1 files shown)


QGarchery:
Hi @aryansalam, thanks for opening. The borrow rate logged in `AccrueInterest` is really the previous borrow rate that it used until this point from the last interest accrual (note that `borrowRate` is not view)

PR #765, opened 2026-03-17
chore: add revert message to FlashBorrowerMock

Ocheretovich:
Add revert message to improve debugging when onMorphoFlashLoan is called by an unexpected sender


QGarchery:
Hi @Ocheretovich, thanks for your time but we don't accept external contributions on this repository

PR #768, opened 2026-04-04
Add SLSA generic generator workflow

rich887687-cmd:


QGarchery:
Thanks for your submission @rich887687-cmd, but we don't accept external contributions

PR #780, opened 2026-07-19
[Certora] remove liquidate buffer

QGarchery:
Rationale is that it doesn't prove what it says it proves. What it actually proves is already proven in `liquidateDecreasesBorrowExchangeRate`.
See [this run](https://prover.certora.com/output/66603/ca2443ca7cc84eaabd286b87bedc87eb?anonymousKey=bc91020d2db7815cb6965fb0096e6774d6c92c53) where the rule in liquidate buffer is simplified, so it looks more like `liquidateDecreasesBorrowExchangeRate`


certora/specs/LiquidateBuffer.spec
```diff
@@ -1,71 +0,0 @@
...
-    uint256 repaidShares = assert_uint256(borrowerShares - newBorrowerShares);
-    uint256 newVirtualTotalAssets = virtualTotalBorrowAssets(id);
-    uint256 newVirtualTotalShares = virtualTotalBorrowShares(id);
-
-    // Hint for the prover to show that there is no bad debt realization.
-    assert newBorrowerCollateral != 0;
-    // Hint for the prover about the ratio used to close the position.
-    assert repaidShares * borrowerCollateral >= seizedAssets * borrowerShares;
-    // Prove that the ratio of shares of debt over collateral is smaller after the liquidation.
-    assert borrowerShares * newBorrowerCollateral >= newBorrowerShares * borrowerCollateral;
```
MathisGD, line 67: curious that we were doing this complex things if it was as simple as `liquidateDecreasesBorrowExchangeRate`

MathisGD (approved):

QGarchery:
I confused the first step for a hint, so when I removed it becomes a redundant spec. The 2 steps are:
- `assert borrowerShares * newBorrowerCollateral >= newBorrowerShares * borrowerCollateral` ← this is not a hint, it's the meat of the proof. You get a better shares/collat ratio
- the "shares ratio improves", so you get a better debt/shares ratio ← this is similar to `liquidateDecreasesBorrowExchangeRate` but needed for the whole proof (better debt/collat ratio)

